Issue #30, opened 2026-06-07
[persistence] barcode_to_accumulated_persistence defaults verbose=True (prints a tqdm progress bar),


bwehlin:
_Found by an automated masspcf API bug scan (2026-06-07); independently reproduced and verified. **Severity: low.**_

### Summary
Documented default contradicts the actual default and produces unexpected console output for a non-interactive call; it is also inconsistent with the two sibling functions in the same module.

### Minimal reproduction
```python
import masspcf as mpcf
import masspcf.persistence as mpers
import numpy as np
import inspect

# Signature inspection
print('AP verbose default =', inspect.signature(mpers.barcode_to_accumulated_persistence).parameters['verbose'].default)  # True
print('SR verbose default =', inspect.signature(mpers.barcode_to_stable_rank).parameters['verbose'].default)             # False
print('BC verbose default =', inspect.signature(mpers.barcode_to_betti_curve).parameters['verbose'].default)            # False

# Runtime: AP prints a tqdm bar to stderr, SR is silent
bcs = mpcf.zeros((3,), dtype=mpcf.barcode64)
for i in range(3):
    bcs[i] = mpers.Barcode(np.array([[0., 1.], [0., 2.]]))
mpers.barcode_to_accumulated_persistence(bcs)  # emits "Converting barcodes to accumulated persistence functions:" tqdm bar on stderr
mpers.barcode_to_stable_rank(bcs)              # silent
# Run from /tmp, not /workspaces/masspcf, to use the installed package.
```

### Observed
inspect.signature reports verbose default = True for barcode_to_accumulated_persistence (vs False for both barcode_to_stable_rank and barcode_to_betti_curve). At runtime, calling barcode_to_accumulated_persistence(bcs) with no verbose argument emits a tqdm progress bar on stderr ("Converting barcodes to accumulated persistence functions: ...0%/100%..."), while the sibling barcode_to_stable_rank(bcs) produces no output. Exit code 0; no crash. Source: /workspaces/masspcf/masspcf/persistence/barcode_summary.py line 113 declares `verbose: bool = True`, while the docstring (lines 137-138) says "Show progress information, by default False". The siblings at lines 57 and 84 default to verbose=False.

### Expected
[...]

PR #102, opened 2026-06-07
fix: barcode_to_accumulated_persistence verbose default should be False

mukejane:
Fixes issue #30. The function signature declared `verbose: bool = True` but the docstring, sibling functions, and Sphinx docs all document `False` as the default. This caused unexpected tqdm progress bar output on default calls.


bwehlin:
Many thanks!

PR #121, opened 2026-06-18
Give every tensor NumPy-correct __bool__ (no __len__ fallback)

### LINKED ISSUE #45: [interop] bool() on FloatTensor/IntTensor falls back to __len__ (length, not value): every non-empty numeric
_Found by an automated masspcf API bug scan (2026-06-07); independently reproduced and verified. **Severity: high.**_

### Summary
NumPy-semantics violation and a silent footgun: `if some_float_tensor:` is always True. Wrong-kind/confusing error ('len() of unsized object') on 0-d.

### Minimal reproduction
```python
# size-1 zero -> should be False (numpy), is True
print('bool(IntTensor([0])) =', bool(mpcf.IntTensor(np.array([0]))))      # True (wrong; numpy False)
print('bool(FloatTensor([0.0])) =', bool(mpcf.FloatTensor(np.array([0.0]))))  # True (wrong; numpy False)
# multi-element -> should raise ValueError (numpy), does not
print('multi:', bool(mpcf.FloatTensor(np.array([0.0,0.0]))))             # True (wrong; numpy raises ValueError)
# 0-d -> wrong-kind error from __len__ fallback
try:
    bool(mpcf.FloatTensor(np.array(0.0)))
except Exception as e:
    print('0-d raised', type(e).__name__, ':', e)                        # TypeError: len() of unsized object
# proof of mechanism: bool tracks length, not value
import masspcf.tensor as mt
print('NumericTensor has __bool__:', '__bool__' in vars(mt.NumericTensor))  # False -> falls back to __len__
print('len([0.0,0.0]) =', len(mpcf.FloatTensor(np.array([0.0,0.0]))))       # 2 -> truthy
```

### Observed
NumericTensor (parent of FloatTensor/IntTensor) defines __float__/__int__ but no __bool__, so Python's bool() falls back to Tensor.__len__ (masspcf/_tensor_base.py:766), which returns the first-dim length. Confirmed from /tmp against the installed package (exit 0, no crash): bool(IntTensor([0])) -> True; bool(FloatTensor([0.0])) -> True; bool(FloatTensor([0.0, 0.0])) -> True (no exception, because len==2 is truthy); bool(FloatTensor(np.array(0.0))) (0-d) -> TypeError: 'len() of unsized object'. So every non-empty numeric tensor is truthy regardless of its values, multi-element tensors silently fail to raise, and 0-d tensors raise a confusing wrong-kind error. BoolTensor.__bool__ (tensor.py:328) already implements the correct semantics; numeric tensors do not inherit it. No test or docstring sanctions this: test_tensor_comparison.py only exercises bool() on BoolTensor (the result of ==).

### Expected
Match NumPy / the codebase's own BoolTensor contract: bool() on a size-1 numeric tensor returns the element's truthiness (bool(IntTensor([0])) -> False, bool(FloatTensor([0.0])) -> False, bool(FloatTensor([5.0])) -> True), and bool() on a tensor with more than one element raises ValueError('The truth value of a tensor with more than one element is ambiguous'). numpy reference confirms: bool(np.array([0])) is False, bool(np.array([0.0,0.0])) raises ValueError. A 0-d tensor should evaluate its single element rather than raising 'len() of unsized object'.

### Root-cause hint
masspcf/tensor.py: NumericTensor lacks __bool__, so bool() falls through to _tensor_base.py:Tensor.__len__ (returns first-dim length, and raises 'len() of unsized object' for 0-d). BoolTensor.__bool__ (tensor.py:328) is the correct reference implementation; lift it to NumericTensor (or a shared base) using size/shape so size-1 returns element truthiness, size!=1 raises ValueError, and 0-d (size==1) works.

```python
# Bug 45: bool() on a numeric tensor reports length, not value truthiness.
def test_bool_numeric_tensor_matches_numpy_truthiness():
    """bool() on FloatTensor/IntTensor must use value truthiness, not __len__."""
    # BUG: bool() on FloatTensor/IntTensor falls back to __len__ (length, not value)
    # Expected: NumPy semantics -- size-1 returns element truthiness; multi-element
    #           raises ValueError("...more than one element is ambiguous"); a 0-d
    #           tensor evaluates its single element instead of raising on len().
    # Observed today: every non-empty numeric tensor is truthy (length>0); a 0-d
    #           tensor raises TypeError("len() of unsized object").

    # size-1 zero -> False (currently True, because len==1 is truthy)
    assert bool(mpcf.IntTensor(np.array([0]))) is False
    assert bool(mpcf.FloatTensor(np.array([0.0]))) is False
    # size-1 nonzero -> True
    assert bool(mpcf.FloatTensor(np.array([5.0]))) is True

    # multi-element -> ValueError like numpy (currently returns True silently)
    with pytest.raises(ValueError, match="more than one element"):
        bool(mpcf.FloatTensor(np.array([0.0, 0.0])))

    # 0-d tensor -> evaluate the single element (currently raises
    # TypeError: len() of unsized object)
    assert bool(mpcf.FloatTensor(np.array(0.0))) is False
    assert bool(mpcf.FloatTensor(np.array(7.0))) is True
```

</details>

bwehlin:
## Summary

`NumericTensor` (`FloatTensor`/`IntTensor`) had no `__bool__`, so `bool()` fell back to `Tensor.__len__` and returned the **first-dimension length** instead of the value:

- `bool(IntTensor([0]))` → `True` (NumPy: `False`)
- `bool(FloatTensor([0.0, 0.0]))` → `True` — no error (NumPy raises `ValueError`)
- `bool(FloatTensor(np.array(0.0)))` (0-d) → `TypeError: len() of unsized object`

So `if some_float_tensor:` was effectively always true — a silent footgun.

## Fix

`BoolTensor` already had the correct implementation. Rather than copy it, I **lifted `__bool__` onto the shared `Tensor` base** and removed the `BoolTensor` duplicate, so numeric, bool, and 0-d tensors all behave identically and like NumPy:

- exactly one element (incl. 0-d scalar) → that element's truthiness;
- any other size → `ValueError` ("ambiguous").

The message now reports the element count, so it is accurate for the empty (0-element) case too (the old `BoolTensor` wording said "more than one element"). One existing test that pinned the old wording is updated to match on `"ambiguous"`.

Docs: updated the "Converting to Python bool" section of `docs/arithmetic.rst`.

## Tests

`test/python/test_bugscan_bool.py`:
- `test_size1_truthiness_matches_value` — `int32/int64/float32/float64`, value `0`/`5`/`-3`
- `test_zero_d_truthiness` — 0-d scalar
- `test_multi_element_bool_raises` — multi-float/int, empty, 2-D → `ValueError`
- `test_numpy_parity`, `test_bool_tensor_still_works` (inheritance), `test_if_statement_reflects_value`

Full suite green: **1644** Python tests pass (31 skipped, CUDA); the lift to the base did not disturb any other tensor type.

Fixes #45

---
_Generated by [Claude Code](https://claude.ai/code/session_01QLUKg7KqfRzMq9A5VLr6pj)_


bwehlin:
## 🔍 Adversarial review

### Summary
The PR gives numeric tensors a correct `__bool__` by lifting it from `BoolTensor` onto the shared `Tensor` base and deleting the duplicate. The core logic (`size == 1` → element truthiness, else `ValueError`) is correct, NumPy-consistent, and the 0-d path is sound (`assert_valid_index` accepts an empty index for rank-0, so `_get_element([])` works). The fix is solid for the numeric/bool case the issue is about. My findings concern the *blast radius* of putting `__bool__` on the base: several object-element tensor types now silently get new `bool()` semantics that are neither tested nor documented.

### 🟡 Edge cases to confirm

**1. The lift to `Tensor` changes truthiness for every object-element tensor, but only numeric/bool types are tested.**
`__bool__` now lives on `Tensor` (`stablebear/_tensor_base.py`, new block ~L781–796), so `PcfTensor`, `IntPcfTensor`, `PointCloudTensor`, `DistanceMatrixTensor`, `SymmetricMatrixTensor`, and `BarcodeTensor` all inherit it for the first time. The PR description says "every tensor," yet `test/python/test_bugscan_bool.py` exercises only `FloatTensor`/`IntTensor`/`BoolTensor`. For a multi-element or 0-d object tensor this *changes* behavior (previously `bool()` fell back to `Tensor.__len__` and returned the first-dim length → truthy with no error; now it raises `ValueError`). That's arguably an improvement, but it is an untested, cross-type behavior change. Worth at least one test over an object tensor (e.g. a multi-element `PcfTensor` → `ValueError`).

**2. Size-1 `BarcodeTensor` truthiness now depends on the barcode's bar count, not on the cell being set.**
The new `__bool__` calls `bool(self._data._get_element(idx))` on the **raw C++ element**. The C++ barcode type binds `__len__` (`src/python/persistence/py_barcode.cpp:73` → `self.bars().size()`), so for a size-1 `BarcodeTensor` holding an *empty* barcode (0 bars), `bool()` now returns `False`, whereas before it returned `True` (via `Tensor.__len__` → `shape[0] == 1`). This is a surprising, type-dependent semantic with no test and no doc note. (By contrast `Pcf`/`PointCloud`/matrix C++ types bind neither `__bool__` nor `__len__`, so they stay truthy — so the behavior is also *inconsistent across* object types.)

### 🟢 Nits

**3. `__bool__` bypasses `_represent_element`.** Unlike `__getitem__` (`_tensor_base.py:379`, which returns `self._represent_element(self._data._get_element(idx))`), the new `__bool__` calls `bool()` directly on the raw C++ element. For numeric tensors this is fine, but it is precisely what makes finding #2 depend on raw C++ object semantics rather than the Python wrapper. If object-tensor truthiness is intended to be supported, consider routing through the wrapper (or explicitly restricting `__bool__` to numeric/bool tensors) and documenting the intent.

**4. Doc wording vs. exception text.** `docs/arithmetic.rst` comments the error as `ValueError: ambiguous truth value`; the actual message is `"The truth value of a tensor with {size} elements is ambiguous. Use array_equal()..."`. Minor, but the doc snippet won't match verbatim. (The message itself is an improvement over the old "more than one element" wording — it now reads correctly for the 0-element case.)

### Verdict
The numeric `__bool__` fix is correct and the tests for it are appropriate. The real gap is coverage/intent for the object-element tensor types that now inherit `__bool__` from the base — most notably the `BarcodeTensor` empty-barcode case (finding #2), which is a genuine, untested behavior change. None of these block the headline fix, but #1 and #2 deserve either a test or an explicit scoping decision before merge.


---

Issue #39, opened 2026-06-07
[matrices] Element assignment into PointCloud/DistanceMatrix/SymmetricMatrix tensors aliases source storage


bwehlin:
_Found by an automated masspcf API bug scan (2026-06-07); independently reproduced and verified. **Severity: high.**_

### Summary
Violates standard assignment-copies semantics that the __setitem__ docstring implies (it documents only view-on-read for basic indexing, never aliasing-on-write of stored elements). The element types (mpcf::PointCloud<T>/DistanceMatrix/SymmetricMatrix) hold a std::shared_ptr buffer; _decay_value returns val._data and _set_element copies the element by value, copying the shared_ptr rather than the data. Leads to silent cross-cell corruption.

### Minimal reproduction
```python
import numpy as np
import masspcf as mpcf
from masspcf.distance_matrix import DistanceMatrix
from masspcf.symmetric_matrix import SymmetricMatrix

# 1) mutate source after assignment corrupts the cell
t = mpcf.zeros((2,), dtype=mpcf.distmat64)
m = DistanceMatrix(3); m[0,1] = 2.0
t[0] = m
m[0,2] = 8.0                      # mutate source AFTER assignment
assert t[0][0,2] == 0.0, f'ALIASED: cell changed to {t[0][0,2]}'  # FAILS: 8.0

# 2) same source into two cells -> cells alias each other (distmat + symmat)
for dt, mk in [(mpcf.distmat64, lambda: DistanceMatrix(3)),
               (mpcf.symmat64, lambda: SymmetricMatrix(3))]:
    tt = mpcf.zeros((2,), dtype=dt)
    shared = mk()
    tt[0] = shared; tt[1] = shared
    tt[0][0,1] = 99.0
    assert tt[1][0,1] == 0.0, f'{dt.__name__} ALIASED: {tt[1][0,1]}'  # FAILS: 99.0

# 3) pcloud: source FloatTensor and sibling cell both aliased
pt = mpcf.zeros((2,), dtype=mpcf.pcloud64)
src = mpcf.FloatTensor(np.zeros((3,2)))
pt[0] = src; pt[1] = src
pt[0][0,0] = 99.0
assert float(pt[1][0,0]) == 0.0   # FAILS: 99.0  (sibling aliased)
assert float(src[0,0]) == 0.0     # FAILS: 99.0  (source aliased)
```

### Observed
[...]

PR #125, opened 2026-06-18
Copy overlapping RHS in assign_from (self-aliasing slice assignment)

### LINKED ISSUE #6: [setitem] Self-aliasing basic-slice assignment (a[:] = a[::-1], a[1:] = a[:-1]) silently corrupts data
_Found by an automated masspcf API bug scan (2026-06-07); independently reproduced and verified. **Severity: high.**_

### Summary
Silent data corruption / wrong numerical result that disagrees with NumPy semantics the docstring claims. Root cause: Tensor<T>::assign_from (include/mpcf/tensor.tpp) does `mpcf::walk(*this, ...) { (*this)(idx) = T(rhs_view(idx)); }` with no overlap detection and no temporary copy of the RHS. When the RHS is a basic-slice view (a view, per the docstring) of the same tensor, source and destination storage overlap and the walk reads back already-overwritten values. Advanced-index RHS (e.g. a[[0,1,2]] = a[[4,3,2]]) is unaffected because advanced getitem returns a copy. This is the documented but unguarded view-aliasing hazard.

### Minimal reproduction
```python
import numpy as np
import masspcf as mpcf
from masspcf.tensor import FloatTensor

# Case 1: in-place reversal
a = FloatTensor(np.array([1., 2., 3., 4.], dtype=np.float32))
a[:] = a[::-1]
print('mpcf reverse:', np.asarray(a).tolist())   # [4.0, 3.0, 3.0, 4.0]  (WRONG)
np_a = np.array([1., 2., 3., 4.], dtype=np.float32); np_a[:] = np_a[::-1]
print('numpy reverse:', np_a.tolist())            # [4.0, 3.0, 2.0, 1.0]
assert np.asarray(a).tolist() == np_a.tolist(), 'ALIASING CORRUPTION'

# Case 2: 2D row shift-down
b = FloatTensor(np.arange(12, dtype=np.float32).reshape(4, 3))
b[1:] = b[:-1]
print('mpcf shift:', np.asarray(b).tolist())      # all rows [0,1,2] (WRONG)
np_b = np.arange(12, dtype=np.float32).reshape(4, 3); np_b[1:] = np_b[:-1]
assert np.array_equal(np.asarray(b), np_b), 'ALIASING CORRUPTION'

# Control: a fresh copy of the RHS produces the correct NumPy result
c = FloatTensor(np.array([1., 2., 3., 4.], dtype=np.float32))
c[:] = c[::-1].copy()
print('mpcf reverse w/ copy:', np.asarray(c).tolist())  # [4.0, 3.0, 2.0, 1.0] (correct)
```

### Observed
CPU: a[:] = a[::-1] on [1,2,3,4] yields [4.0, 3.0, 3.0, 4.0] instead of [4.0, 3.0, 2.0, 1.0]. b[1:] = b[:-1] on the (4,3) arange tensor yields four identical rows [[0,1,2],[0,1,2],[0,1,2],[0,1,2]] instead of [[0,1,2],[0,1,2],[3,4,5],[6,7,8]]. 1D forward-overlap a[1:]=a[:-1] on arange(6) yields all zeros [0,0,0,0,0,0] instead of [0,0,1,2,3,4]. No exception is raised; the wrong values are written silently. Using a[:] = a[::-1].copy() (a fresh non-aliased RHS) gives the correct result, confirming the cause is in-place read-back of overlapping storage. GPU was not independently verifiable on this host (no NVIDIA GPU; CUDA backend falls back to CPU), but the assignment path (Tensor<T>::assign_from in tensor.tpp) is device-agnostic walk logic, so the same defect applies on both backends.

### Expected
NumPy-equivalent, well-defined results: a[:] = a[::-1] -> [4.0, 3.0, 2.0, 1.0]; b[1:] = b[:-1] -> [[0,1,2],[0,1,2],[3,4,5],[6,7,8]]; arange(6)[1:]=...[:-1] -> [0,0,1,2,3,4]. The __setitem__ docstring promises NumPy-style indexing assignment and __getitem__ documents basic-slice views that share storage; NumPy materializes an overlapping RHS into a temporary before assigning, so the result is well-defined and the library must match it.

### Root-cause hint
include/mpcf/tensor.tpp: Tensor<T>::assign_from (lines 143-158) does mpcf::walk(*this, ...) { (*this)(idx) = T(rhs_view(idx)); } with no overlap/aliasing detection and no temporary copy of the RHS; when the RHS is a basic-slice view of the same storage, the destination walk overwrites elements that the source view later reads back, producing corrupted output. Fix: detect storage overlap between target and rhs_view and materialize the RHS into a temporary before the walk (NumPy semantics).

In `test/python/test_bugscan_setitem.py` -- run with `cd test && python -m pytest python/test_bugscan_setitem.py`

```python
def test_aliased_basic_slice_assignment_matches_numpy():
    """Self-aliasing basic-slice assignment must match NumPy (no in-place corruption)."""
    # BUG: Self-aliasing basic-slice assignment (a[:] = a[::-1], a[1:] = a[:-1])
    #      silently corrupts data because the RHS view shares storage with the
    #      destination and is read back mid-write.
    # Expected: NumPy-equivalent results (NumPy materializes an overlapping RHS
    #           into a temporary before assigning, so the result is well-defined).
    # Observed today: in-place read-back of overlapping storage writes wrong
    #                 values silently (no exception raised).

    # Case 1: in-place reversal.
    a = FloatTensor(np.array([1.0, 2.0, 3.0, 4.0], dtype=np.float32))
    a[:] = a[::-1]
    np_a = np.array([1.0, 2.0, 3.0, 4.0], dtype=np.float32)
    np_a[:] = np_a[::-1]
    assert np.asarray(a).tolist() == np_a.tolist()  # expected [4, 3, 2, 1]

    # Case 2: 2D row shift-down with backward overlap.
    b = FloatTensor(np.arange(12, dtype=np.float32).reshape(4, 3))
    b[1:] = b[:-1]
    np_b = np.arange(12, dtype=np.float32).reshape(4, 3)
    np_b[1:] = np_b[:-1]
    assert np.array_equal(np.asarray(b), np_b)  # expected rows [0,1,2],[0,1,2],[3,4,5],[6,7,8]

    # Case 3: 1D forward overlap.
    c = FloatTensor(np.arange(6, dtype=np.float32))
    c[1:] = c[:-1]
    np_c = np.arange(6, dtype=np.float32)
    np_c[1:] = np_c[:-1]
    assert np.asarray(c).tolist() == np_c.tolist()  # expected [0,0,1,2,3,4]
```

</details>

bwehlin:
## Summary

`Tensor::assign_from` walked the destination element-by-element (`(*this)(idx) = rhs_view(idx)`) with no overlap handling. When the RHS is a view of the **same** tensor, source and destination storage overlap, so the walk read back values it had already overwritten:

```python
a = FloatTensor(np.array([1., 2., 3., 4.]))
a[:] = a[::-1]      # was: [4, 3, 3, 4]   (numpy: [4, 3, 2, 1])
b[1:] = b[:-1]      # was: all rows equal to row 0
```

## Fix

When the RHS shares the destination's **base buffer**, materialize an independent copy of the RHS before the walk — matching NumPy, which copies an overlapping RHS into a temporary. Notes:
- Aliasing is only possible when element types match, so the check is under `if constexpr (std::is_same_v<T, U>)` (cross-dtype assignment, e.g. int→float, can't alias).
- `data()` includes the view offset, so `a[1:]` and `a[:-1]` have *different* `data()` pointers despite sharing a buffer — the base is compared via `data() - offset()`.
- Non-overlapping and cross-dtype assignments keep the original copy-free path.

## Tests

`test/python/test_bugscan_setitem.py`, all asserted against NumPy:
- `test_reverse_in_place_1d`, `test_shift_down_2d`, `test_forward_overlap_1d`
- `test_int_tensor_self_alias`, `test_copy_rhs_control_still_correct`
- `test_non_aliased_assignment_unchanged` (source untouched, copy-free path), `test_partial_overlap_disjoint_halves`

Full suites green: **404** C++ and **1644** Python tests pass (31 skipped).

Fixes #6

---
_Generated by [Claude Code](https://claude.ai/code/session_01QLUKg7KqfRzMq9A5VLr6pj)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Bug fixes
 
 * **A single point cloud is now subscriptable** — a 0-d `PointCloudTensor` (one cloud, e.g. `sb.PointCloudTensor(arr)` from an `(n_points, dim)` array) can be indexed as its underlying array, so the natural plotting idiom `pc[:, 0]` / `pc[:, 1]` works directly instead of raising `IndexError`. Indexing tensors of rank ≥ 1 still selects clouds as before. ([#133](https://github.com/kthtda/stablebear/issues/133))
+* **Self-aliasing slice assignment no longer corrupts data** — assigning an overlapping view of a tensor into itself (`a[:] = a[::-1]`, `a[1:] = a[:-1]`) silently produced wrong results, because the element-wise copy read positions it had already overwritten. The right-hand side is now materialized first when it overlaps the destination (matching NumPy), so `a[:] = a[::-1]` reverses correctly; non-overlapping assignments are unaffected. ([#6](https://github.com/kthtda/stablebear/issues/6))
 
 ## 0.4.3
 
--- a/include/sbear/tensor.tpp
+++ b/include/sbear/tensor.tpp
@@ -136,6 +136,26 @@ namespace sb
     }
 
     auto rhs_view = rhs.broadcast_to(target);
+
+    // If the RHS aliases this tensor's storage (e.g. a[:] = a[::-1] or
+    // a[1:] = a[:-1], where both sides are views of the same buffer), an
+    // element-wise walk would read positions it has already overwritten.
+    // Materialize an independent copy of the RHS first, matching NumPy's
+    // handling of overlapping assignment. Aliasing is only possible when the
+    // element types match and the two views share the same base buffer; data()
+    // includes the view offset, so compare the buffer base (data() - offset()).
+    if constexpr (std::is_same_v<T, U>)
+    {
+      if (this->data() - this->offset() == rhs.data() - rhs.offset())
+      {
+        Tensor<T> rhs_copy = rhs_view.copy();
+        sb::walk(*this, [this, &rhs_copy](const std::vector<size_t>& idx){
+          (*this)(idx) = rhs_copy(idx);
+        });
+        return;
+      }
+    }
+
     sb::walk(*this, [this, &rhs_view](const std::vector<size_t>& idx){
       (*this)(idx) = T(rhs_view(idx));
     });
--- a/test/python/test_bugscan_setitem.py
+++ b/test/python/test_bugscan_setitem.py
@@ -1,0 +1,64 @@
+import numpy as np
+
+import stablebear as sb
+
+
+# ---------------------------------------------------------------------------
+# Bug #6: self-aliasing basic-slice assignment (a[:] = a[::-1], a[1:] = a[:-1])
+# silently corrupted data, because Tensor::assign_from walked element-by-element
+# with no overlap handling and read back values it had already overwritten.
+# NumPy materializes an overlapping RHS into a temporary; match that.
+# ---------------------------------------------------------------------------
+
+
+def test_reverse_in_place_1d():
+    a = sb.FloatTensor(np.array([1.0, 2.0, 3.0, 4.0], dtype=np.float32))
+    a[:] = a[::-1]
+    na = np.array([1.0, 2.0, 3.0, 4.0], dtype=np.float32)
+    na[:] = na[::-1]
+    assert np.asarray(a).tolist() == na.tolist() == [4.0, 3.0, 2.0, 1.0]
+
+
+def test_shift_down_2d():
+    b = sb.FloatTensor(np.arange(12, dtype=np.float32).reshape(4, 3))
+    b[1:] = b[:-1]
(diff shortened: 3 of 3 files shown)


bwehlin:
Folding this into #130, which already contains this commit (the self-aliasing `assign_from` fix and its tests) on top of the object-element overlap fix (#129). #130 carries the combined changelog entries for #6, #39, and #129. Closing in favor of #130.

PR #132, opened 2026-06-20
Normalize negative shape axes

### LINKED ISSUE #17: [shapeops] transpose() rejects negative axes with a raw pybind TypeError; swapaxes normalizes them and NumPy

### Summary
Documented NumPy-like behavior is violated: __getitem__/swapaxes resolve negatives, the docstrings promise NumPy-style indexing, but transpose silently lacks negative-axis support and surfaces a misleading binding-level error. Inconsistent with swapaxes which normalizes the same negative inputs.

### Minimal reproduction
```python
import numpy as np
import masspcf as mpcf

t = mpcf.FloatTensor(np.arange(24, dtype=np.float64).reshape(2, 3, 4))
# valid negative-axis permutation NumPy accepts:
print("numpy [-1,-2,-3]:", np.arange(24).reshape(2,3,4).transpose([-1,-2,-3]).shape)  # (4, 3, 2)
try:
    print("mpcf  [-1,-2,-3]:", tuple(t.transpose([-1,-2,-3]).shape))
except Exception as e:
    print("mpcf  [-1,-2,-3] FAILED:", type(e).__name__, "-", str(e).splitlines()[0])
# swapaxes normalizes the same negatives fine -> inconsistency:
print("mpcf swapaxes(-1,-3):", tuple(t.swapaxes(-1, -3).shape))  # (4, 3, 2)
```

### Observed
t.transpose([-1,-2,-3]) raises "TypeError: transpose(): incompatible function arguments. The following argument types are supported: ... axes: Sequence[SupportsInt|SupportsIndex]". The same TypeError occurs for the valid mixed case t.transpose([2,1,-3]). The error fires at the pybind type-conversion boundary before any of mpcf's own axis validation runs. Meanwhile t.swapaxes(-1,-3) accepts the identical negative axes and returns shape (4,3,2), and NumPy's transpose([-1,-2,-3]) also returns (4,3,2). Note: the reporter's [2,1,-1] example is a poor one -- NumPy itself rejects it (ValueError: repeated axis in transpose, since -1 resolves to 2); the real defect is the valid permutations like [-1,-2,-3] and [2,1,-3]. Also note transpose has no docstring (__doc__ is None), so this is not strictly a docstring-mismatch.

### Expected
transpose should accept negative axes and resolve them against ndim like NumPy (and like swapaxes already does in the same wrapper), so transpose([-1,-2,-3]) returns shape (4,3,2). For genuinely invalid axes it should raise a clear ValueError/IndexError naming the axis (the C++ layer already does this for out-of-range/repeated positive axes), not a raw pybind "incompatible function arguments" TypeError.

### Root-cause hint
masspcf/_tensor_base.py:transpose (line 696) passes list(axes) straight to the C++ binding without resolving negatives; the binding (src/python/py_tensor.hpp:246 -> tensor.hpp:228 Tensor::transpose(const std::vector<size_t>&)) takes unsigned size_t, so pybind rejects negative ints before validation. Fix: normalize negatives against self.ndim in the Python wrapper, mirroring swapaxes (_tensor_base.py:702-711).

In `test/python/test_bugscan_shapeops.py` -- run with `cd test && python -m pytest python/test_bugscan_shapeops.py`

```python
def test_transpose_accepts_negative_axes():
    """transpose() must resolve negative axes like NumPy and swapaxes, not raise a raw TypeError."""
    # BUG: transpose() passes negative axes straight to a size_t C++ binding,
    # so pybind rejects them with "incompatible function arguments".
    # Expected: transpose([-1,-2,-3]) -> shape (4,3,2) (NumPy parity); mixed
    # valid permutations like [2,1,-3] also work.
    # Observed today: raw pybind TypeError before any validation runs.
    t = mpcf.FloatTensor(np.arange(24, dtype=np.float64).reshape(2, 3, 4))

    assert tuple(t.transpose([-1, -2, -3]).shape) == (4, 3, 2)
    assert tuple(t.transpose([2, 1, -3]).shape) == (4, 3, 2)
    # Sanity: swapaxes already resolves the same negatives (the inconsistency).
    assert tuple(t.swapaxes(-1, -3).shape) == (4, 3, 2)
```

</details>

### LINKED ISSUE #18: [shapeops] Tensor.squeeze(axis) rejects negative axes with a raw pybind TypeError instead of resolving them

### Summary
Violates the library's stated NumPy-like axis semantics (negative axes accepted elsewhere) and gives a misleading binding-level error message rather than the documented ValueError.

### Minimal reproduction
```python
import numpy as np
import masspcf as mpcf

ts = mpcf.FloatTensor(np.arange(6, dtype=np.float64).reshape(1, 6, 1))
print('numpy squeeze(-1):', np.arange(6).reshape(1,6,1).squeeze(-1).shape)  # (1,6)
print('numpy squeeze(-3):', np.arange(6).reshape(1,6,1).squeeze(-3).shape)  # (6,1)
try:
    print('mpcf squeeze(-1):', tuple(ts.squeeze(-1).shape))
except Exception as e:
    print('mpcf squeeze(-1) FAILED with', type(e).__name__, ':', str(e).splitlines()[0])
# positive axes work and match numpy:
print('mpcf squeeze(0):', tuple(ts.squeeze(0).shape))  # (6,1)
print('mpcf squeeze(2):', tuple(ts.squeeze(2).shape))  # (1,6)
```

### Observed
ts.squeeze(-1) and ts.squeeze(-3) raise "TypeError: squeeze(): incompatible function arguments. The following argument types are supported: ... axis: SupportsInt|SupportsIndex". This is a raw pybind11 overload-resolution failure (the negative Python int cannot convert to the binding's unsigned size_t parameter), not a library-level error. Positive axes work correctly: squeeze(0) -> (6,1), squeeze(2) -> (1,6), squeeze(None) -> (6,). Behavior is identical across all tensor types since they share the same base method.

### Expected
Per NumPy parity (np.arange(6).reshape(1,6,1).squeeze(-1) -> (1,6); squeeze(-3) -> (6,1)), squeeze should resolve the negative axis against ndim and squeeze it, or at minimum raise a clear IndexError/ValueError. This is the documented/intended idiom: the sibling method swapaxes in the same file resolves negatives and raises IndexError on out-of-range, and test_negative_indexing.py's module docstring explicitly states the suite targets NumPy-parity for negative axes/indices.

### Root-cause hint
masspcf/_tensor_base.py:713 TensorBase.squeeze passes axis straight to the C++ binding without resolving negatives (unlike swapaxes at lines 702-711); the binding src/python/py_tensor.hpp:249 is typed `size_t axis`, so a negative int fails pybind unsigned conversion with a misleading TypeError. Fix: resolve `axis += ndim` when negative and bounds-check in the Python wrapper (mirroring swapaxes).

In `test/python/test_bugscan_shapeops.py` -- run with `cd test && python -m pytest python/test_bugscan_shapeops.py`

```python
def test_squeeze_accepts_negative_axis():
    """squeeze(axis) must resolve a negative axis like NumPy, not raise a raw TypeError."""
    # BUG: squeeze(axis) passes a negative axis to a size_t C++ binding, so
    # pybind rejects it with "incompatible function arguments".
    # Expected (NumPy parity): squeeze(-1) -> (1,6); squeeze(-3) -> (6,1).
    # Observed today: raw pybind TypeError before any validation runs.
    ts = mpcf.FloatTensor(np.arange(6, dtype=np.float64).reshape(1, 6, 1))

    assert tuple(ts.squeeze(-1).shape) == (1, 6)
    assert tuple(ts.squeeze(-3).shape) == (6, 1)
```

</details>

KairosOps:
## Summary

This normalizes negative axes in the Python tensor wrapper before the values cross into the pybind layer.

`transpose()` now resolves each supplied axis against `self.ndim`, and `squeeze(axis)` does the same for a single axis. That keeps these shape operations consistent with NumPy and with the existing `swapaxes()` wrapper, which already handled negative axes before calling the C++ backend.

Fixes #17.
Fixes #18.

## Validation

- `BUILD_WITH_CUDA=0 SKIP_STUBGEN=1 python -m pip install .`
- `cd test && python -m pytest python/test_tensor_reshape.py`
- `git diff --check`

I also ran `python -m ruff check stablebear/_tensor_base.py test/python/test_tensor_reshape.py`. It still reports existing style findings in this area, including local import ordering in `_tensor_base.py` and the unchanged import block in `test_tensor_reshape.py`, so I left those unrelated cleanups out of this bug fix.

### DIFF
--- a/stablebear/_tensor_base.py
+++ b/stablebear/_tensor_base.py
@@ -147,6 +147,16 @@ def _resolve_negative_indices(index_tensor, axis_size):
     return IntTensor(arr)
 
 
+def _resolve_axis(axis: int, ndim: int) -> int:
+    axis = operator.index(axis)
+    resolved = axis + ndim if axis < 0 else axis
+    if not 0 <= resolved < ndim:
+        raise IndexError(
+            f"axis {axis} is out of range for tensor with {ndim} dimension(s)"
+        )
+    return resolved
+
+
 class Tensor(ABC):
     _data: CppTensor
 
@@ -709,7 +719,10 @@ class Tensor(ABC):
         return self._to_py_tensor(self._data.reshape(list(shape)))
 
     def transpose(self, axes=None):
-        return self._to_py_tensor(self._data.transpose(list(axes) if axes else []))
+        if axes is None:
+            return self._to_py_tensor(self._data.transpose([]))
+        resolved_axes = [_resolve_axis(axis, self.ndim) for axis in axes]
+        return self._to_py_tensor(self._data.transpose(resolved_axes))
 
     @property
     def T(self):
@@ -729,7 +742,7 @@ class Tensor(ABC):
     def squeeze(self, axis=None):
         if axis is None:
             return self._to_py_tensor(self._data.squeeze())
-        return self._to_py_tensor(self._data.squeeze(axis))
+        return self._to_py_tensor(self._data.squeeze(_resolve_axis(axis, self.ndim)))
 
     def expand_dims(self, axis):
         return self._to_py_tensor(self._data.expand_dims(axis))
--- a/stablebear/_tensor_base.py
+++ b/stablebear/_tensor_base.py
@@ -729,14 +729,8 @@ class Tensor(ABC):
         return self.transpose()
 
     def swapaxes(self, axis1, axis2):
-        n = self.ndim
-        if axis1 < 0:
-            axis1 += n
-        if axis2 < 0:
-            axis2 += n
-        if not (0 <= axis1 < n and 0 <= axis2 < n):
-            raise IndexError(
-                f"swapaxes: axis out of range for tensor with {n} dimensions")
+        axis1 = _resolve_axis(axis1, self.ndim)
+        axis2 = _resolve_axis(axis2, self.ndim)
         return self._to_py_tensor(self._data.swapaxes(axis1, axis2))
 
     def squeeze(self, axis=None):
@@ -745,6 +739,7 @@ class Tensor(ABC):
         return self._to_py_tensor(self._data.squeeze(_resolve_axis(axis, self.ndim)))
 
     def expand_dims(self, axis):
+        axis = _resolve_axis(axis, self.ndim + 1)
         return self._to_py_tensor(self._data.expand_dims(axis))
 
     # Implementation: looks up a C++ cast function by naming convention:
--- a/test/python/test_tensor_reshape.py
+++ b/test/python/test_tensor_reshape.py
@@ -323,6 +323,9 @@ class TestExpandDims:
     def test_expand_negative_axis(self, TensorType, np_dtype):
         _assert_expand_dims(np.arange(12, dtype=np_dtype).reshape(3, 4), -2, TensorType)
 
+    def test_expand_negative_axis_before_first(self, TensorType, np_dtype):
+        _assert_expand_dims(np.arange(12, dtype=np_dtype).reshape(3, 4), -3, TensorType)
+
     def test_expand_is_view(self, TensorType, np_dtype):
         np_arr = np.arange(6, dtype=np_dtype)
         t = TensorType(np_arr)
@@ -332,8 +335,10 @@ class TestExpandDims:
 
     def test_expand_out_of_range_raises(self, TensorType, np_dtype):
         t = TensorType(np.arange(6, dtype=np_dtype))
-        with pytest.raises((ValueError, RuntimeError)):
+        with pytest.raises((ValueError, RuntimeError, IndexError)):
             t.expand_dims(3)
+        with pytest.raises((ValueError, RuntimeError, IndexError)):
+            t.expand_dims(-3)
 
 
 # --- astype ---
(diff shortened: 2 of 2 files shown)


bwehlin:
Hi, thank you for your contribution! 

In `swapaxes` (`stablebear/_tensor_base.py:702-711`), we have logic that is very similar to `_resolve_axis` added in this PR. I think the version in this PR looks a bit more robust. Could you replace the inline logic in `swapaxes` to use `_resolve_axis` twice (once per axis)? For consistency, we should probably also add this to `expand_dims` rather than resolving this case specifically in C++.

KairosOps:
Thanks for the pointer. I pushed a follow-up that routes both `swapaxes` axes through `_resolve_axis` and normalizes `expand_dims` on the Python side with `self.ndim + 1`, so the valid insertion range stays consistent with NumPy.

Checked locally:
- `BUILD_WITH_CUDA=0 python -m pip install -e .`
- `python -m pytest test/python/test_tensor_reshape.py::TestSwapaxes test/python/test_tensor_reshape.py::TestExpandDims -q`
- `python -m pytest test/python/test_tensor_reshape.py -q`
- `git diff --check`

bwehlin:
Thanks, I'm making a release now and this will go into the new version (0.4.3).

PR #135, opened 2026-06-26
Preserve length-1 batch axis in persistence functions (#28, #29)

### LINKED ISSUE #28: [persistence] compute_persistent_homology squeezes the batch dimension for a genuine length-1 (1,)

### Summary
The squeeze cannot distinguish a genuine multi-element tensor that happens to have leading dim 1 from the wrapped single-cloud convenience path; both reach the same out=zeros((1,)) -> reshaped (1, max_dim+1) state. This directly contradicts the documented output-shape guarantee and silently breaks batch pipelines whenever the batch size is 1.

### Minimal reproduction
```python
import masspcf as mpcf
import masspcf.persistence as mpers
import numpy as np

def run(N):
    X = mpcf.zeros((N,), dtype=mpcf.pcloud64)
    for i in range(N):
        X[i] = np.array([[0.,0.],[1.,0.],[0.,1.]])
    bcs = mpers.compute_persistent_homology(X, max_dim=1, verbose=False)
    print(f'N={N}: output shape={bcs.shape} (docstring guarantees ({N}, 2))')
    try:
        bcs[0, 0]
        print('  bcs[0,0] OK')
    except Exception as e:
        print('  bcs[0,0] FAILED:', type(e).__name__, e)

run(2)  # shape (2, 2), bcs[0,0] works
run(1)  # shape (2,) -- batch dim dropped; bcs[0,0] -> IndexError

# Same defect on the distance-matrix path:
def make_dm():
    d = mpcf.DistanceMatrix(3, dtype=mpcf.float64)
    d[0,1]=1.0; d[0,2]=1.0; d[1,2]=np.sqrt(2)
    return d
for N in (1, 2):
    dmt = mpcf.zeros((N,), dtype=mpcf.distmat64)
    for i in range(N):
        dmt[i] = make_dm()
    bcs = mpers.compute_persistent_homology(dmt, max_dim=1, verbose=False)
    print(f'DM N={N}: output shape={bcs.shape}')
```

### Observed
For a genuine 1-D input tensor of shape (1,) -- both PointCloudTensor and DistanceMatrixTensor -- compute_persistent_homology returns a BarcodeTensor of shape (max_dim+1,) (e.g. (2,) for max_dim=1) instead of (1, max_dim+1). Consequently bcs[0, 0] raises "IndexError: too many indices for tensor: tensor is 1-dimensional, but 2 were indexed". N=2 gives (2,2) and bcs[0,0] works; N=3 gives (3,2); shapes (1,1),(1,3),(2,1) are all correct -- the defect is uniquely the leading-and-only-dimension==1 case. Confirmed for both backends' Python path (device=both; CPU verified, code path is device-independent). Exit code 0 (no crash), the failure is the wrong shape plus a downstream IndexError on natural batch indexing.

### Expected
Per the docstring (homology.py lines 67-68: "For an input tensor of shape (d_1, ..., d_n), the output has shape (d_1, ..., d_n, max_dim+1)") and the established batch convention pinned by test_pcloud_tensor_single_point_per_cloud (input (3,) -> output (3,2) with bcs[i,0] indexing), a genuine tensor input of shape (1,) must yield output shape (1, max_dim+1) = (1, 2), so bcs[0,0] works. The trailing squeeze should only collapse the scalar-convenience inputs (single np.ndarray, single FloatTensor, single DistanceMatrix), which the existing tests document as returning shape (max_dim+1,).

### Root-cause hint
masspcf/persistence/homology.py:compute_persistent_homology -- the unconditional squeeze "if len(out.shape)==2 and out.shape[0]==1: out = out[0,:]" at lines 123-124 (distmat path) and 157-158 (pcloud path) cannot distinguish a wrapped scalar-convenience input (np.ndarray/FloatTensor/single DistanceMatrix, all wrapped into a (1,) tensor at lines 105-112 / 129-137) from a user-supplied genuine (1,)-shaped batch tensor; both reach out=zeros((1,))->reshaped (1,max_dim+1). Fix: track a "was_scalar_input" flag at wrap time and squeeze only when set, instead of inferring it from out.shape[0]==1.

In `test/python/test_bugscan_persistence.py` -- run with `cd test && python -m pytest python/test_bugscan_persistence.py`

```python
def test_compute_persistent_homology_preserves_length1_batch_dim():
    """Bug 27: genuine (1,) tensor input is squeezed to (max_dim+1,) not (1, max_dim+1)."""
    # BUG: compute_persistent_homology squeezes the batch dimension for a genuine
    #      length-1 (1,) PointCloudTensor/DistanceMatrixTensor.
    # Expected: per the docstring "For an input tensor of shape (d_1, ..., d_n),
    #           the output has shape (d_1, ..., d_n, max_dim+1)", a genuine
    #           (1,)-shaped batch tensor must yield output shape (1, max_dim+1)
    #           and bcs[0, 0] must index the first element's H0.
    # Observed today: output shape (max_dim+1,) = (2,); bcs[0, 0] -> IndexError.

    # Point-cloud path.
    X = mpcf.zeros((1,), dtype=mpcf.pcloud64)
    X[0] = np.array([[0.0, 0.0], [1.0, 0.0], [0.0, 1.0]])
    bcs = mpers.compute_persistent_homology(X, max_dim=1, verbose=False)
    assert bcs.shape == (1, 2), (
        f"genuine (1,) pcloud batch should give shape (1, 2), got {bcs.shape}")
    # Natural batch indexing must work (this raises IndexError today).
    bcs[0, 0]

    # Distance-matrix path: same defect.
    d = mpcf.DistanceMatrix(3, dtype=mpcf.float64)
    d[0, 1] = 1.0
    d[0, 2] = 1.0
    d[1, 2] = np.sqrt(2.0)
    dmt = mpcf.zeros((1,), dtype=mpcf.distmat64)
    dmt[0] = d
    bcs_dm = mpers.compute_persistent_homology(dmt, max_dim=1, verbose=False)
    assert bcs_dm.shape == (1, 2), (
        f"genuine (1,) distmat batch should give shape (1, 2), got {bcs_dm.shape}")
    bcs_dm[0, 0]
```

</details>

### LINKED ISSUE #29: [persistence] barcode_to_stable_rank/betti_curve/accumulated_persistence drop the leading dimension for any 2-D

### Summary
The squeeze in barcode_summary.py lines 50-51 ('if len(out.shape) == 2 and out.shape[0] == 1: out = out[0, :]') fires for any 2-D output whose first axis is 1, not only the single-Barcode convenience case. Note the single-Barcode path returns a Pcf (not a Tensor) and the 1-D BarcodeTensor case correctly stays 1-D, so this squeeze is overly broad and contradicts the documented shape contract.

### Minimal reproduction
```python
import masspcf as mpcf
import masspcf.persistence as mpers
import numpy as np

bcs = mpcf.zeros((1, 2), dtype=mpcf.barcode64)
bcs[0, 0] = mpers.Barcode(np.array([[0., 1.]]))
bcs[0, 1] = mpers.Barcode(np.array([[0., 2.]]))
out = mpers.barcode_to_stable_rank(bcs)
print('stable_rank (1,2) ->', out.shape)   # Shape(2)  -- should be (1,2)
try:
    out[0, 1]
except Exception as e:
    print('out[0,1] FAILED:', type(e).__name__, e)  # IndexError: too many indices

out_betti = mpers.barcode_to_betti_curve(bcs)
print('betti (1,2) ->', out_betti.shape)    # Shape(2)
ap = mpers.barcode_to_accumulated_persistence(bcs, verbose=False)
print('apf (1,2) ->', ap.shape)             # Shape(2)

bcs3 = mpcf.zeros((1, 3), dtype=mpcf.barcode64)
for j in range(3):
    bcs3[0, j] = mpers.Barcode(np.array([[0., 1. + j]]))
print('stable_rank (1,3) ->', mpers.barcode_to_stable_rank(bcs3).shape)  # Shape(3)

# Proof the C++ backend is NOT at fault: it writes the correct shape; Python squeezes it.
from masspcf import _mpcf_cpp as cpp
from masspcf.async_task import _run_task
backend = cpp.persistence.PersistenceBarcodeSummary64
raw = mpcf.zeros((1,), dtype=mpcf.pcf64)
_run_task(lambda: backend.spawn_barcode_to_stable_rank_task(bcs._data, raw._data), verbose=False)
print('raw C++ output shape for (1,2) input:', raw.shape)  # Shape(1, 2) -- correct

# Controls (all preserved correctly):
for shp in [(2,2),(2,1),(1,2,2)]:
    import itertools
    b = mpcf.zeros(shp if (shp:=shp) else shp, dtype=mpcf.barcode64)
    for idx in itertools.product(*[range(s) for s in shp]):
        b[idx] = mpers.Barcode(np.array([[0., 1.+sum(idx)]]))
    print(shp, '->', mpers.barcode_to_stable_rank(b).shape)
```

### Observed
All three summary functions (barcode_to_stable_rank, barcode_to_betti_curve, barcode_to_accumulated_persistence) return a PcfTensor whose leading dimension has been dropped whenever the input BarcodeTensor is 2-D with leading dim 1. Verified outputs: input (1,2) -> output Shape(2); input (1,3) -> Shape(3); input (1,1) -> Shape(1). Indexing the (2,)-shaped result with out[0,1] raises "IndexError: too many indices for tensor: tensor is 1-dimensional, but 2 were indexed". The C++ backend itself writes the CORRECT shape (verified: spawn_barcode_to_stable_rank_task writes Shape(1,2) for a (1,2) input); the dimension is dropped purely by the Python wrapper at barcode_summary.py lines 50-51. Controls behave correctly: (2,2)->(2,2), (2,1)->(2,1), (2,2,...) and 3-D leading-1 shapes like (1,2,2)->(1,2,2), and the single-Barcode path returns a Pcf via a separate branch. No existing test in test/python/persistence covers the (1,N) 2-D case.

### Expected
Per each function's docstring ("a PcfTensor with the same shape as the input" for non-single inputs), input shape (1,2) must yield output shape (1,2) and (1,3) must yield (1,3). The shape must be preserved exactly, and out[0,1] must succeed. The squeeze should not fire for legitimate multi-element (1,N) BarcodeTensors; the single-Barcode convenience case is already handled separately (returns a Pcf via the isinstance(X, Barcode) branch), so the squeeze serves no documented purpose.

### Root-cause hint
masspcf/persistence/barcode_summary.py:_barcode_to_pcf lines 50-51 -- the guard `if len(out.shape) == 2 and out.shape[0] == 1: out = out[0, :]` is overly broad: it squeezes every 2-D output whose first axis is 1, dropping a real dimension for genuine (1,N) tensors. The C++ task already produces the correct (1,N) shape, and the single-Barcode case returns a Pcf via a separate earlier branch, so this squeeze is unnecessary and should be removed (or scoped so it never collapses a legitimate multi-element tensor).

In `test/python/test_bugscan_persistence.py` -- run with `cd test && python -m pytest python/test_bugscan_persistence.py`

```python
def test_barcode_summaries_preserve_leading_dim1_shape():
    """Bug 28: summary funcs drop the leading dim for any 2-D (1, N) BarcodeTensor."""
    # BUG: barcode_to_stable_rank/betti_curve/accumulated_persistence drop the
    #      leading dimension for any 2-D BarcodeTensor with leading dim 1
    #      (e.g. (1,2) -> (2,)).
    # Expected: per each function's docstring ("a PcfTensor with the same shape
    #           as the input"), input (1, 2) must yield output (1, 2) and (1, 3)
    #           must yield (1, 3); out[0, 1] must succeed.
    # Observed today: output is squeezed to (2,) / (3,); out[0, 1] -> IndexError.

    bcs = mpcf.zeros((1, 2), dtype=mpcf.barcode64)
    bcs[0, 0] = mpers.Barcode(np.array([[0.0, 1.0]]))
    bcs[0, 1] = mpers.Barcode(np.array([[0.0, 2.0]]))

    sr = mpers.barcode_to_stable_rank(bcs)
    assert sr.shape == (1, 2), f"stable_rank should preserve (1, 2), got {sr.shape}"
    sr[0, 1]  # must not raise

    betti = mpers.barcode_to_betti_curve(bcs)
    assert betti.shape == (1, 2), f"betti_curve should preserve (1, 2), got {betti.shape}"
    betti[0, 1]

    apf = mpers.barcode_to_accumulated_persistence(bcs, verbose=False)
    assert apf.shape == (1, 2), (
        f"accumulated_persistence should preserve (1, 2), got {apf.shape}")
    apf[0, 1]

    # Also pin (1, 3) -> (1, 3) for stable_rank to confirm it is not special to N=2.
    bcs3 = mpcf.zeros((1, 3), dtype=mpcf.barcode64)
    for j in range(3):
        bcs3[0, j] = mpers.Barcode(np.array([[0.0, 1.0 + j]]))
    sr3 = mpers.barcode_to_stable_rank(bcs3)
    assert sr3.shape == (1, 3), f"stable_rank should preserve (1, 3), got {sr3.shape}"
```

</details>

bwehlin:
## Summary

Two persistence functions dropped a genuine length-1 leading axis, collapsing a one-element *batch* into an unbatched result:

- **#28** `compute_persistent_homology` on a `PointCloudTensor` / `DistanceMatrixTensor` of shape `(1,)` returned shape `(max_dim+1,)` instead of `(1, max_dim+1)`.
- **#29** `barcode_to_stable_rank` / `barcode_to_betti_curve` / `barcode_to_accumulated_persistence` on a `(1, N)` `BarcodeTensor` returned `(N,)` instead of `(1, N)`.

Both used `if len(out.shape) == 2 and out.shape[0] == 1: out = out[0, :]`, which cannot tell a real one-element batch from the scalar-convenience case.

## Fix

- `homology.py`: track a `was_scalar_input` flag, set only on the scalar-convenience branches (single `ndarray` / `FloatTensor` / `DistanceMatrix`), and drop the leading axis only then. Genuine `PointCloudTensor` / `DistanceMatrixTensor` batches keep their shape.
- `barcode_summary.py`: remove the squeeze entirely — the single-`Barcode` convenience case already returns a `Pcf` from an earlier branch, and the backend writes a result of the same shape as the input tensor.

The docstrings already documented the (now-true) shape contract, so no doc change was needed.

## Tests

Added regression tests in `test/python/test_bugscan_persistence.py`:
- genuine `(1,)` point-cloud and distance-matrix batches keep `(1, 2)`, and the elements stay 2-D-indexable;
- multi-element batches `(2,)` are unchanged;
- scalar-convenience inputs (ndarray / `FloatTensor` / `DistanceMatrix`) still return `(2,)`;
- all three barcode summaries preserve `(1, 2)` and still return a `Pcf` for a single `Barcode`.

All persistence tests pass (`pytest python/persistence python/test_bugscan_persistence.py` → 62 passed).

Closes #28
Closes #29


bwehlin:
## 🔍 Adversarial review

### Summary
This PR fixes two genuine shape bugs (#28, #29) where persistence functions squeezed a real length-1 *batch* axis, conflating it with the scalar-convenience case. I traced the fix against the actual C++ backend and adjacent tests and tried hard to break it; **the change is correct, well-targeted, and the tests genuinely exercise the fix.** Verdict: looks good to merge.

### Verification (why this is sound)
The fix relies on the backend reshaping `out` to mirror the input batch shape, and that mechanism is confirmed in the C++:

- `compute_persistent_homology` → `RipserTaskImpl::run_async` in `include/sbear/persistence/compute_persistence.hpp:163-165` reassigns the output: `shape = m_input.shape(); shape.emplace_back(maxDim+1); m_ret = Tensor(shape)`. So a genuine `(1,)` batch yields `(1, maxDim+1)` and a single (scalar-convenience) input — wrapped into a `(1,)` tensor — yields the same `(1, maxDim+1)`. The old `if len(out.shape)==2 and out.shape[0]==1` squeeze could not tell these apart; the new `was_scalar_input` flag (`homology.py:90-148`) sets the squeeze **only** on the `DistanceMatrix` / `FloatTensor` / `ndarray` branches, which is exactly right.
- `barcode_to_*` → `BarcodeSummaryTask::run_async` in `include/sbear/persistence/barcode_summary.hpp:37` does `m_ret = Tensor(m_barcodes.shape())`, i.e. output shape == input shape. Removing the squeeze in `barcode_summary.py:35-38` is therefore correct, and the single-`Barcode` convenience case returns a `Pcf` earlier at `barcode_summary.py:25-26` (a `Barcode` passes through `_get_backend` via its `.dtype` attribute), so it never reaches the removed squeeze.

### No regressions in adjacent tests (checked explicitly)
- `test/python/persistence/test_persistence_edge_cases.py:83` (`bcs.shape == (1,)`, `max_dim=0`) — input is a plain `np.ndarray` → scalar-convenience → still squeezed `(1,1)→(1,)`. Passes.
- `test/python/test_integration.py:29` (`bcs.shape == (2,)`) — input `dm` is a single `DistanceMatrix` from `sb.pdist` on a one-cloud `(8,)` input (its `to_dense()` is a single 2-D matrix) → scalar-convenience → still squeezed to `(2,)`. Passes.
- Multi-element and higher-rank batches (`test_ripser.py` `(3,2,7,4)`, `(4,)`) were already ≥2 trailing dims / not `(1,*)`, so unaffected.

### Test quality
The new tests in `test/python/test_bugscan_persistence.py` actually distinguish the two cases the bug conflated:
- genuine `(1,)` point-cloud and dist-matrix batches assert `(1, 2)` *and* index `out[0,0]` — fails under the old code;
- scalar-convenience inputs (ndarray / `FloatTensor` / `DistanceMatrix`) assert `(2,)`;
- barcode summaries build a genuine `(1,2)` via `[0:1]` slicing (self-validated before the call) and assert `(1,2)` plus a single-`Barcode` `Pcf` return.

All referenced factory APIs exist (`DistanceMatrix.from_dense`, `DistanceMatrixTensor.from_numpy`, `mp.Barcode`, `sb.Pcf`), so the tests won't error out spuriously. The `mp.Barcode(np.array(...))` default-`float64` path constructs a `Barcode64` and hits the single-`Barcode` branch as intended.

### Findings
No likely bugs and no edge cases that break. A couple of minor, non-blocking observations:

### 🟢 Nits
- The shape contract is now carried implicitly by the C++ backend (`m_ret = Tensor(input.shape())`). Since the whole bug class came from a Python-side squeeze guessing the backend's intent, a one-line comment in `homology.py` / `barcode_summary.py` pointing at `RipserTaskImpl`/`BarcodeSummaryTask` as the source of truth for output shape would help future readers — the existing comments are good but don't name the backend reshape.
- Not introduced here, but worth noting: 0-d (rank-0) `BarcodeTensor` / point-cloud-tensor inputs to these functions return rank-1 / 0-d results that are neither squeezed nor flagged. Behavior is unchanged by this PR (so not a regression), but there's no test pinning it; a follow-up could lock it down if 0-d batches are a supported input.

Overall: solid, minimal, and the tests are real regression tests rather than tautologies. 👍


---

PR #136, opened 2026-06-26
Clean errors for invalid setitem dtypes on cloud/matrix tensors (#41, #42)

### LINKED ISSUE #41: [matrices] PointCloudTensor advertises float/int as valid setitem dtypes but scalar assignment passes

### Summary
Inconsistency between the advertised valid setitem dtypes and what _decay_value can actually handle; the __setitem__ docstring promises a TypeError 'If val has a type that cannot be assigned to this tensor', but here a supposedly-valid type slips past validation and produces a confusing constructor error instead.

### Minimal reproduction
```python
import numpy as np
import masspcf as mpcf

t = mpcf.zeros((2,), dtype=mpcf.pcloud64)
valid = t._get_valid_setitem_dtypes()
assert float in valid and int in valid, valid  # advertised as valid
for v in [3.14, 7, np.float64(3.0)]:
    try:
        t[0] = v
        print('assigned', repr(v), '-> shape', t[0].shape)
    except TypeError as e:
        print('assign', repr(v), 'raised:', e)
# Output:
# assign 3.14 raised: Cannot create FloatTensor from <class 'float'>
# assign 7 raised: Cannot create FloatTensor from <class 'int'>
# assign np.float64(3.0) raised: Cannot create FloatTensor from <class 'numpy.float64'>
# t[:] = 1.0 fails identically. Exit code 0 (clean TypeError, not a hard crash).
```

### Observed
PointCloudTensor._get_valid_setitem_dtypes() returns [FloatTensor, np.ndarray, float, int], so float/int (and np.float64, which subclasses float) all PASS _validate_setitem_dtype. Assignment then enters _decay_value, which calls FloatTensor(val, dtype=...). FloatTensor.__init__ only accepts cpp.*Tensor / FloatTensor / np.ndarray / list / tuple, so a bare scalar raises `TypeError: Cannot create FloatTensor from <class 'float'>` (resp. int / numpy.float64). The same happens for t[0]=3.14, t[0]=7, t[:]=1.0, and t[0]=np.float64(3.0). The error references FloatTensor, a different class than the PointCloudTensor being assigned to, making it misleading. For a real NumericTensor, scalar fill works (e.g. ft[:] = 1.0 succeeds), so the validator's advertisement is inconsistent with actual behavior only for PointCloudTensor. Confirmed: a point-cloud cell is an (n, dim) array (test_point_cloud_tensors.py assigns np.random.randn(n, dim)); scalars are not meaningful RHS values for such cells.

### Expected
Internal consistency between advertised valid setitem dtypes and what _decay_value can handle.

### Root-cause hint
masspcf/tensor.py:PointCloudTensor._get_valid_setitem_dtypes (line 295-296) lists float/int as valid, but PointCloudTensor._decay_value (line 290-293) calls FloatTensor(val, ...) which FloatTensor.__init__ (line 122-140) rejects for bare scalars; validator and _decay_value disagree about supported RHS types.

```python
# Bug 40: pcloud advertises float/int as valid setitem dtypes but scalar
#         assignment slips past validation and raises a misleading
#         "Cannot create FloatTensor from <class float>".
def test_pcloud_scalar_assignment_raises_clean_consistent_error():
    """Advertised pcloud setitem dtypes must match what assignment accepts."""
    # BUG: PointCloudTensor._get_valid_setitem_dtypes() advertises float/int,
    #      so scalar assignment passes validation, then _decay_value calls
    #      FloatTensor(scalar) which raises a misleading constructor error.
    # Expected: internal consistency -- a scalar RHS either assigns cleanly or is
    #           rejected by validation with a Python-level message that does NOT
    #           reference the unrelated FloatTensor constructor.
    # Observed today: TypeError "Cannot create FloatTensor from <class 'float'>"
    #                 (resp. int / numpy.float64) -- a leaked internal message.

    t = mpcf.zeros((2,), dtype=mpcf.pcloud64)
    valid = t._get_valid_setitem_dtypes()

    for v in (3.14, 7, np.float64(3.0)):
        if float in valid or int in valid:
            # The validator advertises scalars as valid, so assignment must
            # actually succeed (consistency between advertisement and behavior).
            tt = mpcf.zeros((2,), dtype=mpcf.pcloud64)
            tt[0] = v  # must not raise
        else:
            # If scalars are NOT advertised, validation must reject them with a
            # clean message that does not leak the FloatTensor constructor error.
            with pytest.raises(TypeError) as excinfo:
                t[0] = v
            assert "Cannot create FloatTensor" not in str(excinfo.value), (
                f"leaked internal constructor error: {excinfo.value}")
```

</details>

### LINKED ISSUE #42: [matrices] DistanceMatrix/SymmetricMatrix tensor element assignment with mismatched dtype leaks a raw pybind

### Summary
Validation only checks the Python wrapper class (DistanceMatrix) not its dtype, so a precision mismatch falls through to the raw C++ _set_element and surfaces a confusing low-level message. Numeric tensors handle cross-precision RHS via _coerce_rhs; matrix-element tensors do not.

### Minimal reproduction
```python
import masspcf as mpcf
from masspcf.distance_matrix import DistanceMatrix

t = mpcf.zeros((2,), dtype=mpcf.distmat64)
dm32 = DistanceMatrix(3, dtype=mpcf.float32); dm32[0,1] = 3.0
try:
    t[0] = dm32
    print('assigned (cast?) dtype:', t[0].dtype)
except TypeError as e:
    print('raised TypeError:', str(e).splitlines()[0])
# Output: raised TypeError: _set_element(): incompatible function arguments. ...
# Invoked with: <...DistanceMatrix64Tensor object ...>, [0], DistanceMatrix_f32(size=3)
```

### Observed
Assigning a float32 DistanceMatrix element into a float64 DistanceMatrixTensor (t[0] = dm32) raises a raw pybind11 TypeError whose message leaks internal C++ type names: "_set_element(): incompatible function arguments. The following argument types are supported: 1. (self: masspcf._mpcf_cuda12.DistanceMatrix64Tensor, arg0: ..., arg1: masspcf._mpcf_cuda12.DistanceMatrix_f64) -> None. Invoked with: <...DistanceMatrix64Tensor object...>, [0], DistanceMatrix_f32(size=3)". The same raw error appears for: f64->f32 element assign, SymmetricMatrix element assign, and the tensor-RHS slice path (which leaks __setitem__() instead). The exception IS a TypeError (so the type is correct), but the message is the raw binding diagnostic, not a Python-level explanation.

### Expected
A clear, Python-level TypeError naming the dtype mismatch, e.g. "cannot assign a float32 DistanceMatrix into a float64 DistanceMatrixTensor", consistent with how numeric tensors handle cross-precision RHS. Confirmed asymmetry: numeric tensors cast cleanly (ft[0] = it[0] casts int32->float64; ft[1] = f32[0] casts f32->f64; ft2[:] = f32 casts a whole tensor), because NumericTensor RHS goes through _coerce_rhs -> astype. Matrix-element tensors have no astype, so the conservative correct behavior is a clean dtype-aware error rather than a silent cast.

### Root-cause hint
masspcf/_tensor_base.py:Tensor._validate_setitem_dtype (with DistanceMatrixTensor/SymmetricMatrixTensor._get_valid_setitem_dtypes in distance_matrix.py / symmetric_matrix.py): validation only does isinstance class checks ([DistanceMatrix, DistanceMatrixTensor]) and ignores element dtype, so a precision mismatch passes and _basic_setitem calls _data._set_element(idx, val._data) with a DistanceMatrix_f32 into a float64 tensor, surfacing the raw pybind error. _coerce_rhs (used for the numeric tensor/tensor-RHS path) is not applied to the element-RHS path and matrix wrappers lack astype.

</details>

bwehlin:
## Summary

Two setitem paths leaked raw internal errors instead of a clear `TypeError`:

- **#41** `PointCloudTensor` advertised `float`/`int` as valid setitem dtypes, so `t[i] = 3.14` passed validation and then raised the misleading `Cannot create FloatTensor from <class 'float'>` from `_decay_value`. A point-cloud cell is an `(n_points, dim)` array, so a scalar is not a meaningful right-hand side.
- **#42** `DistanceMatrixTensor` / `SymmetricMatrixTensor` accepted any element or tensor RHS of the right *class* regardless of 32/64-bit precision, leaking the raw pybind `_set_element(): incompatible function arguments` (element RHS) or `__setitem__(): ...` (tensor-slice RHS) dump on a mismatch.

## Fix

- `PointCloudTensor._get_valid_setitem_dtypes` no longer lists `float`/`int`, so a scalar is rejected up front with the standard validator message.
- A shared `_check_matrix_tensor_setitem` helper (in `distance_matrix.py`, reused by both matrix tensor classes via a `_validate_setitem_dtype` override) checks the 32/64-bit precision after the base type check and raises a `TypeError` naming both dtypes. The matrix wrappers store an exact element type and have no astype cast, so this errors rather than silently casting (matching NumPy object-array assignment).

Both are pure Python interop fixes; valid assignments (array / `FloatTensor` into a cloud, matching-precision element / tensor into a matrix tensor) are unchanged.

## Tests

New `test/python/test_bugscan_setitem_dtype.py` (18 cases): scalar-into-cloud raises a clean error across pcloud32/64 and float/int/np.float64; array-into-cloud still works; mismatched matrix element/tensor precision raises a `TypeError` naming the dtypes and *not* containing the pybind text, across both matrix families and both directions; matching-precision assignment still works.

Full Python suite passes (1817 passed).

Closes #41
Closes #42


bwehlin:
## 🔍 Adversarial review

### Summary
The PR turns two leaky internal errors into clean `TypeError`s: scalar-into-`PointCloudTensor` (#41, by dropping `float`/`int` from the cloud's valid setitem dtypes) and mismatched 32/64-bit precision when assigning a `DistanceMatrix`/`SymmetricMatrix` element or whole tensor into a matrix tensor (#42, via a shared `_check_matrix_tensor_setitem` helper wired into a `_validate_setitem_dtype` override). I tried hard to break it and could not find a correctness defect — the change is small, well-scoped, and the tests genuinely exercise the claimed behavior (including negative assertions that the pybind text is gone). **Verdict: solid, recommend merge.**

I could not build the C++ extension here (submodules not vendored in the review env), so the findings below are from static reasoning against the actual code on the head ref and `main`, not a live test run.

### Findings

#### 🟢 Nits / observations (no action strictly required)

- **Validation runs before any mutation — confirmed safe.** `Tensor.__setitem__` (`_tensor_base.py:508`) calls `_validate_setitem_dtype(val)` before `_setitem_entries`, so the new precision check rejects a mismatch with no partial write, on both the single-index element path and the `dst[:] = src` slice path. Good.

- **Internal construction is not broken by the new strict check.** `_matrix_tensor_cpp_from_array` (`distance_matrix.py:207-212`) pre-casts the source array to `np_float` matching the target `dt` before `t[idx] = scalar_cls.from_dense(...)`, so the element it constructs always matches `self.dtype` and passes the new check. Likewise `_pointcloud_cpp_from_list` assigns `np.ascontiguousarray(...)` (an `ndarray`), still a valid RHS. So the tightened validators don't regress the constructors that drive them.

- **`np.float64(3.0)` is correctly rejected** even though it's an `isinstance(..., float)` — it isn't an `ndarray`, and `float` was removed, so it fails up front. The test covers exactly this. Nice.

- **Intentional asymmetry worth a one-line doc note.** Cloud assignment *recasts* a mismatched-precision RHS: `PointCloudTensor._decay_value` does `FloatTensor(val, dtype=float_dtype)` (`base_tensor.py:379-382`), so a `float32` `FloatTensor` (or `float32` ndarray) assigned into a `pcloud64` cell is silently upcast to f64, whereas the matrix path now *rejects* a `float32` element into a `distmat64` tensor. Both are defensible (the PR comment explains matrices "store an exact element type and do not silently cast"), but the cloud side does the opposite. If matching NumPy object-array semantics is the goal across the board, the cloud recast is the odd one out. Not a bug — just calling out the inconsistency for the record.

- **Lazy vs module-level import of the helper.** `symmetric_matrix.py` imports `_check_matrix_tensor_setitem` inside the method (to avoid a circular import with `distance_matrix`), while `distance_matrix.py` calls it at module scope. Works fine; a top-level import in a shared module would be marginally cleaner, but the current form is acceptable.

### Verdict
No likely bugs and no must-fix edge cases. The fix does what its description claims, the tests assert the right things (both the new clean message *and* the absence of the raw pybind dump, across both matrix families and both mismatch directions), and adjacent behavior (matching-precision assignment, array-into-cloud, internal tensor construction) is preserved. 👍

---

Issue #133, opened 2026-06-22
[friction/matrices] A single PointCloud element is not subscriptable, blocking the natural pc[:, 0] / pc[:, 1] plotting idiom


bwehlin:
_Usability friction. Not a correctness bug._  **Priority: medium.**

### Scenario
A user has a **single point cloud** (one element extracted from a `PointCloudTensor`, or a 0-d cloud) and wants to scatter-plot it via the natural NumPy idiom:

```python
plt.scatter(pc[:, 0], pc[:, 1])
```

### Current approach
The point cloud is not subscriptable, so `pc[:, 0]` / `pc[:, 1]` raise. The user must first round-trip through NumPy:

```python
arr = np.asarray(pc)
plt.scatter(arr[:, 0], arr[:, 1])
```

### Desired API
Make a single point cloud subscriptable so coordinate/column slicing (`pc[:, 0]`, `pc[:, 1]`, `pc[i]`) works directly, matching the NumPy idiom users reach for when plotting.

### Notes
Closely related to #93 (no `to_dense()` / `np.asarray()` round-trip *for the equal-shape PointCloudTensor case*). That issue covers converting a tensor of clouds back to a dense ndarray; this one is specifically about a **single** cloud element not supporting `[:, col]` slicing. Consider whether to fold into #93 or address together.

PR #137, opened 2026-06-26
Read a fresh PointCloudTensor cell back as an empty (0,2) cloud (#44)

### LINKED ISSUE #44: [matrices] zeros(dtype=pcloud*) fresh cell reads back as 0-d scalar 0.0, not the documented "empty point cloud"

### Summary
Doc says zeros gives 'an empty point cloud' but the default-constructed element is a 0-d tensor holding 0.0, not an empty 2-D cloud, so shape semantics are inconsistent between a default cell and an assigned cell. Low severity / arguably a documentation vs behavior mismatch.

### Minimal reproduction
```python
import numpy as np
import masspcf as mpcf

t = mpcf.zeros((1,), dtype=mpcf.pcloud64)
el = t[0]
print('fresh empty cloud: shape', el.shape, 'ndim', el.ndim, 'value', np.asarray(el))
# -> shape Shape() ndim 0 value 0.0   (a 0-d scalar, NOT an empty cloud)

t[0] = np.zeros((0, 2))
print('after assign (0,2): shape', t[0].shape)   # -> Shape(0, 2)

# sibling dtypes honor their "empty/zero collection" docstrings on a fresh cell:
print('symmat fresh:', repr(np.asarray(mpcf.zeros((1,), dtype=mpcf.symmat64)[0])))   # size=0 (0x0)
print('distmat fresh:', repr(np.asarray(mpcf.zeros((1,), dtype=mpcf.distmat64)[0]))) # size=0 (0x0)
print('barcode fresh:', repr(np.asarray(mpcf.zeros((1,), dtype=mpcf.barcode64)[0]))) # shape (0,2)
# only pcloud returns a 0-d scalar 0.0

# downstream consumer is unaffected (no crash, correct result):
import masspcf.persistence as ph
print('ripser on default-empty pcloud:', ph.compute_persistent_homology(mpcf.zeros((2,), dtype=mpcf.pcloud64), max_dim=1).shape)  # Shape(2,2), same as explicit (0,2)
```

### Observed
mpcf.zeros((1,), dtype=mpcf.pcloud64)[0] returns a FloatTensor with shape Shape() (ndim 0) whose value is the scalar 0.0 (np.asarray -> array(0.)). After t[0] = np.zeros((0,2)) the same cell reads back as Shape(0, 2). The two representations differ and t2[0].array_equal(np.zeros((0,2))) is False for a fresh cell. Point cloud is the only one of the four "empty/zero collection" dtypes whose fresh cell does NOT read back as an empty collection: symmat/distmat fresh cells are 0x0 matrices (size=0) and barcode fresh cells are (0,2) empty arrays, all matching their docstrings. The library's own downstream consumer (compute_persistent_homology / ripser) treats a default-empty pcloud cell identically to an explicitly-assigned (0,2) cloud -- no crash and identical results -- so there is no observable functional breakage; the defect is purely the Python read-back representation contradicting the documented contract.

### Expected
Per the zeros() docstring (masspcf/tensor_create.py:59 "dtype=pcloud32/64: An empty point cloud."), echoed in docs/tensors.rst:26 and docs/concepts.rst:129, a freshly-zeroed pcloud cell should read back as an empty point cloud (a 2-D empty array, consistent with the (0, dim) representation an assigned empty cloud produces and with how the sibling collection dtypes symmat/distmat/barcode behave), not as a 0-d scalar holding 0.0.

### Root-cause hint
PointCloudTensor._represent_element (masspcf/tensor.py:287) returns the raw default-constructed inner tensor as the cell; that inner element comes from the C++ default ctor Tensor() : Tensor({}, {}) (include/mpcf/tensor.hpp:81), which builds a 0-d (shape {}) scalar tensor holding T{} = 0.0 rather than a (0, dim) empty cloud -- so the "empty point cloud" the docstring promises is actually a scalar zero. Unlike symmat/distmat/barcode, pcloud has no element wrapper to normalize the empty representation.

```python
# Bug 43: zeros(dtype=pcloud*) fresh cell reads back as a 0-d scalar 0.0 instead
#         of the documented "empty point cloud".
def test_pcloud_fresh_cell_is_empty_cloud_not_scalar():
    """A freshly-zeroed pcloud cell must read back as an empty 2-D cloud."""
    # BUG: zeros(dtype=pcloud*) cells default to a 0-d scalar 0.0 rather than the
    #      documented empty point cloud.
    # Expected: per the zeros() docstring ("an empty point cloud") and the sibling
    #           collection dtypes (symmat/distmat 0x0, barcode (0,2)), a fresh
    #           pcloud cell reads back as a 2-D empty array, matching what an
    #           explicitly-assigned empty cloud produces.
    # Observed today: shape Shape() (ndim 0), value scalar 0.0.

    t = mpcf.zeros((1,), dtype=mpcf.pcloud64)
    fresh = t[0]

    # An assigned empty (0, dim) cloud is the reference representation.
    t2 = mpcf.zeros((1,), dtype=mpcf.pcloud64)
    t2[0] = np.zeros((0, 2))

    assert fresh.ndim == 2, f"fresh pcloud cell has ndim {fresh.ndim}, expected 2"
    assert tuple(fresh.shape)[0] == 0, (
        f"fresh pcloud cell is not empty: shape {tuple(fresh.shape)}")
    assert fresh.array_equal(t2[0]), (
        "fresh pcloud cell does not match an explicitly-assigned empty cloud")
```

</details>

bwehlin:
## Summary

A never-assigned cell of `zeros(dtype=pcloud32/pcloud64)` read back as a 0-d scalar `0.0` (the C++ default-constructed `Tensor`) instead of the documented "empty point cloud":

```python
sb.zeros((1,), dtype=sb.pcloud64)[0]   # was shape () / value 0.0
```

The sibling dtypes already normalize their fresh cells (`distmat`/`symmat` → `(0, 0)`, `barcode` → `(0, 2)`); point clouds were the outlier.

## Fix

`PointCloudTensor._represent_element` normalizes a 0-d default cell to an empty `(0, 2)` cloud of the matching float precision. A real cloud is always 2-D, so the `ndim == 0` sentinel never masks assigned data, and the 0-d single-cloud indexing path (`pc[:, 0]`) is unaffected.

The docstrings already promised an "empty point cloud", so the fix makes behavior match the existing docs (no doc change needed).

## Tests

Added to `test/python/test_point_cloud_tensors.py`: a fresh `pcloud32`/`pcloud64` cell reads back as `(0, 2)` with the right float dtype and compares equal to an explicitly-assigned empty cloud; a real cloud assigned into one cell does not change the still-empty sibling cell.

`test_point_cloud_tensors.py`, `test_numpy_constructors.py`, `test_bugscan_pcloud_setitem.py`, and the persistence suite pass.

Closes #44


bwehlin:
## 🔍 Adversarial review

_Automated adversarial review._

### Summary

The PR makes a never-assigned `pcloud32`/`pcloud64` cell read back as an empty `(0, 2)` cloud instead of a 0-d scalar `0.0`, by normalizing the 0-d default cell in `PointCloudTensor._represent_element`. I tried to break it and the change holds up: the root cause and the sentinel are both correct, and the new tests genuinely exercise the fix. **Verdict: looks correct and safe to merge.** A couple of edge cases below are worth a glance but none are blocking.

I was unable to build/run the package in this environment (the `3rd/` submodules — pybind11, googletest, taskflow — can't be fetched through the proxy), so the review is static. To compensate I traced the relevant C++.

### What I verified

- **Root cause is real.** `include/sbear/tensor.hpp:391` has `using PointCloud = Tensor<T>;`, so a point-cloud element literally *is* a `Tensor<float>`. A fresh cell is a default-constructed `Tensor<float>`, which is 0-d / scalar `0.0` — exactly what #44 reports. The sibling dtypes (`DistanceMatrix`/`SymmetricMatrix`/`Barcode`) use dedicated element types that default-construct to empty matrices/barcodes, so they were never affected. The "pcloud was the outlier" framing is accurate.
- **The `ndim == 0` sentinel cannot mask assigned data.** `_pointcloud_cpp_from_array` enforces `cloud_ndim >= 1` (`base_tensor.py:281`), so every *assigned* cloud is ≥1-D. An empty assigned cloud is still `(0, 2)` → `ndim == 2`, so it falls through and is returned unchanged. Confirmed no masking.
- **The test's `array_equal` on two `(0, 2)` clouds is meaningful, not vacuous.** `Tensor::operator==` (`tensor.tpp:52`) checks shape equality first, then `walk`s; for matching `(0, 2)` shapes the walk visits nothing and returns `true`. So `fresh.array_equal(t[1])` after assigning an explicit empty cloud genuinely passes (and would fail if shapes differed). The dtype assertion (`np.asarray(fresh).dtype == np_float`) is backed by building the sentinel with `dtype=_PCLOUD_TO_FLOAT_DTYPE[self.dtype]`.
- **`test_fresh_pcloud_cell_does_not_mask_real_cloud` is a real regression guard** — it assigns a `(3, 2)` cloud to `t[0]` and asserts `t[1]` is still `(0, 2)`, exercising both the "don't mask" and "sibling still empty" paths.

### 🟡 Edge cases to confirm

- **The normalized shape is hardcoded `(0, 2)` regardless of `cloud_ndim`.** `PointCloudTensor` doesn't retain the `cloud_ndim` used to build it, so for a tensor created with `cloud_ndim != 2` (e.g. the `(30, 2, 20)` / `(40, 15, 10)` 3-D clouds already in `test_can_create_point_clouds`, or a 1-D `cloud_ndim=1` build), a *fresh* cell reads back as 2-D `(0, 2)` while *assigned* cells are 3-D/1-D. This is a minor self-inconsistency, not a correctness bug — `(0, 2)` is a defensible default and matches the docstring. Worth a one-line comment noting the assumption, or leaving as-is.

- **The 0-d single-cloud + fresh-cell path is untested.** `_represent_element` also feeds `_single_cloud()` (`base_tensor.py:361`), which backs the `pc[:, 0]` idiom from #133. For `sb.zeros((), dtype=pcloud64)` (a 0-d tensor whose lone cell was never assigned), the change now makes `pc[:, 0]` return an empty column instead of raising `IndexError` — an *improvement* and consistent with this fix, but the new tests only cover rank-1 (`zeros((2,))`). Consider adding a case asserting a fresh 0-d cloud is `(0, 2)` and `pc[:, 0]` is empty, to lock in that interaction.

### 🟢 Nits

- `_represent_element` always constructs `FloatTensor(element)` first, then on the 0-d branch discards it and builds a second `FloatTensor(np.zeros((0, 2)), ...)`. Negligible cost (the discarded wrap is over a 0-d tensor), but you could check `element.ndim` (or wrap lazily) to avoid the throwaway object. Purely cosmetic.

Overall: small, well-scoped fix that matches the existing docs and the sibling-dtype behavior, with tests that actually assert the right thing. Nice.

---

PR #138, opened 2026-06-26
Dtype-aware default bounds for iterate_rectangles on integer PCFs (#13)

### LINKED ISSUE #13: [pcf] iterate_rectangles fails on integer PCFs with default args: a=0.0/b=float('inf') invalid for the

### Summary
_BACKEND_MAP explicitly includes cpp.Pcf_i32_i32 and cpp.Pcf_i64_i64, advertising int support, and the docstring places no dtype restriction. A user with int PCFs calling the function the documented way gets an opaque pybind TypeError rather than rectangles or a clear error.

### Minimal reproduction
```python
from masspcf.functional import Pcf
fi = Pcf(np.array([[0, 3], [1, 2]], dtype=np.int32))
gi = Pcf(np.array([[0, 2], [1, 4]], dtype=np.int32))
try:
    r = mpcf.iterate_rectangles(fi, gi)        # default a=0.0, b=inf
    print('OK:', [(x.left, x.right) for x in r])
except TypeError as e:
    print('BUG:', str(e).splitlines()[0])
# Works only when BOTH bounds passed as Python ints:
r2 = mpcf.iterate_rectangles(fi, gi, a=0, b=10)
print('explicit int a,b OK:', [(x.left, x.right, x.f_value, x.g_value) for x in r2])
# Partial arg still fails (a stays float 0.0):
try:
    mpcf.iterate_rectangles(fi, gi, b=10)
    print('b-only OK')
except TypeError as e:
    print('b-only BUG:', str(e).splitlines()[0])
# No way to express infinity for an int PCF:
try:
    mpcf.iterate_rectangles(fi, gi, a=0, b=float('inf'))
    print('b=inf OK')
except TypeError as e:
    print('b=inf BUG:', str(e).splitlines()[0])
```

### Observed
For two valid int32 PCFs, mpcf.iterate_rectangles(fi, gi) with default arguments raises "TypeError: iterate_rectangles(): incompatible function arguments. The following argument types are supported: ...". The same failure occurs when passing only b=10 (a defaults to float 0.0) and when passing b=float('inf') (no int representation). It only succeeds when BOTH a and b are passed as Python ints (e.g. a=0, b=10), which works correctly and returns int-valued rectangles [(0,1,3,2),(1,10,2,4)]. Float PCFs work fine with the same defaults. int32/int64 are advertised public dtypes (mpcf.int32, mpcf.int64, IntPcfTensor) and are listed in _BACKEND_MAP (Pcf_i32_i32, Pcf_i64_i64).

### Expected
iterate_rectangles should work out of the box for int PCFs, matching float behavior. The function advertises int support via _BACKEND_MAP and its docstring imposes no dtype restriction, so the documented default call iterate_rectangles(f, g) should succeed for every supported dtype. The fix needs dtype-appropriate defaults: a=0 cast to the time type and b chosen so "to infinity" is expressible for int PCFs (e.g. type-max, or handle +inf/None in C++/Python), rather than hard-coded float defaults that pybind11 cannot bind to the int-typed C++ overload.

### Root-cause hint
masspcf/functional/pcf.py:iterate_rectangles (line 359) hard-codes float defaults a=0.0, b=float('inf'); for int PCFs _BACKEND_MAP routes to the C++ overload Backend<Tt,Tv>::iterate_rectangles (src/python/functional/py_pcf.cpp:108) whose a/b are Tt-typed (int), so pybind11 rejects the float defaults and there is no way to express infinity for an int time axis.

In `test/python/test_bugscan_pcf.py` -- run with `cd test && python -m pytest python/test_bugscan_pcf.py`

```python
# Bug 12: iterate_rectangles fails on integer PCFs with default arguments.
def test_iterate_rectangles_int_pcf_default_args_succeeds():
    """iterate_rectangles must work out of the box for int PCFs (advertised dtype).

    int32/int64 are advertised public dtypes routed through ``_BACKEND_MAP``, and
    the docstring imposes no dtype restriction, so the documented default call
    ``iterate_rectangles(f, g)`` should succeed for them. The Python wrapper
    hard-codes float defaults (``a=0.0``, ``b=float('inf')``) that pybind11
    cannot bind to the int-typed C++ overload.
    """
    # BUG: iterate_rectangles hard-codes float defaults a=0.0/b=inf; int PCFs
    #      route to an int-typed C++ overload that rejects the float defaults.
    # Expected: default-arg call on int PCFs returns the correct int rectangles,
    #           matching the explicit-int-bounds result.
    # Observed today: TypeError "incompatible function arguments" on the default
    #                 call (and on b-only / b=inf), only explicit int a,b works.
    fi = Pcf(np.array([[0, 3], [1, 2]], dtype=np.int32))
    gi = Pcf(np.array([[0, 2], [1, 4]], dtype=np.int32))

    # Reference: the documented/working result with explicit int bounds.
    ref = mpcf.iterate_rectangles(fi, gi, a=0, b=10)
    ref_tuples = [(x.left, x.right, x.f_value, x.g_value) for x in ref]
    assert ref_tuples == [(0, 1, 3, 2), (1, 10, 2, 4)]

    # The default-arg call must SUCCEED (today it raises TypeError before
    # producing anything). The exact int representation of "to infinity" for
    # the trailing rectangle's `right` is implementation-defined, so assert only
    # the structurally-determined parts: two rectangles, matching f/g step
    # values and left edges, with the last one extending past its left edge.
    r = mpcf.iterate_rectangles(fi, gi)
    got = [(x.left, x.right, x.f_value, x.g_value) for x in r]
    assert len(got) == 2
    assert got[0] == (0, 1, 3, 2)
    left, right, fv, gv = got[1]
    assert (left, fv, gv) == (1, 2, 4)
    assert right > left
```

</details>

bwehlin:
## Summary

`iterate_rectangles(f, g)` on integer-typed PCFs raised a raw pybind `TypeError: incompatible function arguments` because the default bounds `a=0.0`, `b=float('inf')` cannot bind to the integer-typed C++ overload. Explicit integer bounds (`iterate_rectangles(f, g, 0, 10)`) already worked.

## Fix

Resolve the default bounds against the PCF's time dtype (Python-only wrapper change, no rebuild):

- **integer PCF** → `a=0`, `b=iinfo(dtype).max` (the integer analogue of "no upper bound"). An explicit `inf` upper bound maps to the dtype maximum; any other explicit bound is truncated to `int`.
- **float PCF** → unchanged (`a=0.0`, `b=inf`), so the existing float behaviour (the last rectangle's `right` is `inf`) is preserved — the float defaults are *not* routed through the C++ `infinite_time()` sentinel, which is the finite `DBL_MAX`.

The docstring is updated (autodoc picks it up in `docs/concepts.rst`).

## Tests

Added integer-PCF cases to `test/python/test_iterate_rectangles.py` (parametrized over int32/int64): default bounds now succeed and the unbounded right edge equals the dtype maximum; explicit integer bounds; explicit `inf` maps to the maximum. The existing float `pcf_dtype` parametrization is unchanged and still asserts the infinite last edge.

`test_iterate_rectangles.py`, `test_pcf_int.py`, `test_pcf.py` all pass.

Closes #13


bwehlin:
## 🔍 Adversarial review

### Summary

The PR makes `iterate_rectangles` default bounds dtype-aware so integer-typed PCFs no longer hit a raw pybind `TypeError` (the float `0.0`/`inf` defaults couldn't bind to the int-typed C++ overload). I cross-checked the change against the C++ kernel, the pybind binding, and the type maps. **The fix is correct and well-targeted, the tests genuinely exercise it, and I found no correctness bugs** — only a couple of minor/edge notes below.

What I verified:
- `src/python/functional/py_pcf.cpp:92,218-221` binds `a, b` as `Tt` (the time type), so float `0.0`/`inf` against an `int32`/`int64` overload is exactly the reported `TypeError`. The fix coerces to `int` for integer PCFs. ✅
- In `include/sbear/algorithms/functional/iterate_rectangles.hpp:85-86`, the last rectangle's `right` is set to `b`, so passing `iinfo.max` (int) / `float('inf')` (float) yields the documented "unbounded right edge" — matching the new and existing tests. ✅
- `f.ttype` ∈ {`int32`,`int64`,`float32`,`float64`} are all keys in `_SB_TO_NP` (`stablebear/typing.py:75`), so `_SB_TO_NP[f.ttype]` is safe. Since PCFs always have `ttype == vtype`, using the time type for the bounds is right. ✅
- `int(iinfo.max)` for int32 (2147483647) / int64 (9223372036854775807) fits the corresponding C++ `int32_t`/`int64_t` parameter exactly — no overflow at the binding. ✅
- The description's claim that the float `inf` default is deliberately *not* routed through the C++ `infinite_time()` sentinel is accurate: `infinity<T>()` returns `std::numeric_limits<T>::max()` (i.e. `DBL_MAX`, finite), confirmed at `include/sbear/functional/time_point.hpp:11-14`. The Python-side `float('inf')` is genuinely infinite and preserves the existing `_assert_rect_inf` behavior. ✅ No regression to the float path.

### 🟢 Nits

- **Doc-location wording (description only, not the code).** The PR description says the updated docstring "autodoc picks it up in `docs/concepts.rst`." `concepts.rst` is hand-written prose with no `autofunction`; the docstring is actually surfaced via `.. automodule:: stablebear.functional.pcf` in `docs/api_base.rst:9`. The code/doc sync is fine — just the description points at the wrong file.

- **Truncation is toward zero, not floor (untested for negatives).** `_coerce_int_bound` uses `int(x)` for finite floats, so `b=-3.7` → `-3` (toward zero), not `-4`. The docstring says "truncated to int," which is consistent, but the truncation test (`test_int_pcf_explicit_float_bound_truncated_to_int`) only covers a positive bound. A negative-bound case would pin down the rounding semantics.

- **`np.float32` inf isn't treated as the sentinel.** `_coerce_int_bound` checks `isinstance(x, float)`, which catches Python `float` and `np.float64` (a `float` subclass) but **not** `np.float32`. So `b=np.float32('inf')` falls to `int(np.float32('inf'))`, raising `OverflowError` instead of mapping to `iinfo.max`. Very unlikely in practice (users pass Python floats), but if you want robustness, gate on `np.isinf(x)` against any float-like rather than `isinstance(x, float)`.

### Verdict

Solid, minimal, correctly-scoped fix with tests that actually hit the previously-broken path (default-bounds call on an integer PCF, asserting the unbounded edge equals the dtype max). No likely bugs or edge-case regressions found; the items above are cosmetic/hardening only.

---

PR #139, opened 2026-06-26
Negative axis/index and dtype-mismatch handling in join/split (#20, #21, #73, #74)

### LINKED ISSUE #20: [joinsplit] concatenate/split/array_split reject negative axis with a pybind "incompatible function arguments"

### Summary
Inconsistent and misleading. stack documents/handles negative axis and the module presents a numpy-like API (tests even cover stack axis=-1), so a user reasonably expects concatenate/split/array_split to accept negative axis too. The error they get instead is a low-level pybind type error that does not mention 'axis' or that negatives are unsupported.

### Minimal reproduction
```python
FT = mpcf.FloatTensor
a = FT(np.arange(6, dtype=np.float64).reshape(2,3))
b = FT(np.arange(6, dtype=np.float64).reshape(2,3))
# stack accepts negative axis (tested at test_tensor_join.py:93):
print('stack axis=-1 ->', np.asarray(mpcf.stack([a,b], axis=-1)).shape)  # (2,3,2)
# concatenate does NOT:
try:
    mpcf.concatenate((a,b), axis=-1)
except Exception as e:
    print('concatenate axis=-1', type(e).__name__, ':', str(e).splitlines()[0])
# contrast: out-of-range POSITIVE axis gives a clean ValueError naming axis:
try:
    mpcf.concatenate((a,b), axis=5)
except Exception as e:
    print('concatenate axis=5', type(e).__name__, ':', str(e).splitlines()[0])
t = FT(np.arange(12, dtype=np.float64).reshape(3,4))
for fn,name in [(lambda: mpcf.split(t,2,axis=-1),'split'),(lambda: mpcf.array_split(t,3,axis=-1),'array_split')]:
    try:
        fn()
    except Exception as e:
        print(name,'axis=-1', type(e).__name__, ':', str(e).splitlines()[0])
# numpy reference (the numpy-like contract this API mirrors):
print('numpy concat axis=-1:', np.concatenate((np.asarray(a),np.asarray(b)),axis=-1).shape)  # (2,6)
```

### Observed
mpcf.concatenate((a,b), axis=-1), mpcf.split(t,2,axis=-1) and mpcf.array_split(t,3,axis=-1) each raise a pybind11 TypeError: "concatenate()/split_sections()/array_split(): incompatible function arguments. The following argument types are supported: ... Invoked with: [...], -1". The error never mentions 'axis' or that negatives are unsupported; it dumps the whole overload signature. By contrast, mpcf.stack([a,b], axis=-1) succeeds and returns shape (2,3,2), and is explicitly covered by an existing test (test/python/test_tensor_join.py:93). Also by contrast, an out-of-range POSITIVE axis (concatenate(..., axis=5)) raises a clean ValueError: "concatenate: axis 5 out of range for 2-D tensor". So negative axis fails in a different, low-level way than other bad axes.

### Expected
For a numpy-like API, negative axis should be accepted and resolved (as stack already does and as numpy does for all three): concatenate((2,3),(2,3), axis=-1) -> (2,6); split((3,4),2,axis=-1) -> two (3,2); array_split((3,4),3,axis=-1) -> (3,2),(3,1),(3,1). At minimum, if negatives are unsupported, the error should be a clear ValueError that names 'axis' (matching the existing positive out-of-range message), not a pybind 'incompatible function arguments' TypeError.

### Root-cause hint
src/python/py_tensor.hpp:251/257/260/263 bind concatenate/split_sections/split_indices/array_split with `size_t axis`, so a negative int fails pybind's unsigned conversion before any C++ logic runs; stack (py_tensor.hpp:254) uses `ptrdiff_t axis` and resolves negatives in include/mpcf/tensor.tpp:978-979. The fix is to give the four functions `ptrdiff_t axis` bindings plus negative-axis resolution in their tensor.tpp implementations (which currently also take `size_t axis`).

In `test/python/test_bugscan_joinsplit.py` -- run with `cd test && python -m pytest python/test_bugscan_joinsplit.py`

```python
# Bug 19: concatenate/split/array_split reject negative axis with a pybind
#         "incompatible function arguments" TypeError, while stack accepts it.
def test_negative_axis_resolves_like_numpy():
    """concatenate/split/array_split must accept negative axis like stack/numpy do."""
    # BUG: negative axis is rejected by a low-level pybind TypeError on
    #      concatenate/split/array_split, even though stack accepts axis=-1
    #      (tested in test_tensor_join.py:93) and the API mirrors numpy.
    # Expected: negative axis is resolved -- concatenate((2,3),(2,3),axis=-1) -> (2,6);
    #      split((3,4),2,axis=-1) -> two (3,2); array_split((3,4),3,axis=-1) ->
    #      (3,2),(3,1),(3,1).  (Weakest acceptable fallback would be a clean
    #      ValueError naming 'axis', but the numpy-like contract is the full result.)
    # Observed today: TypeError "concatenate()/split_sections()/array_split():
    #      incompatible function arguments ...".
    FT = mpcf.FloatTensor

    # concatenate along the last axis.
    a = FT(np.arange(6, dtype=np.float64).reshape(2, 3))
    b = FT(np.arange(6, 12, dtype=np.float64).reshape(2, 3))
    got = np.asarray(mpcf.concatenate((a, b), axis=-1))
    expected = np.concatenate(
        (np.arange(6).reshape(2, 3), np.arange(6, 12).reshape(2, 3)), axis=-1)
    np.testing.assert_array_equal(got, expected)
    assert got.shape == (2, 6)

    # split along the last axis.
    t = FT(np.arange(12, dtype=np.float64).reshape(3, 4))
    parts = mpcf.split(t, 2, axis=-1)
    np_parts = np.split(np.arange(12).reshape(3, 4), 2, axis=-1)
    assert len(parts) == len(np_parts)
    for part, np_part in zip(parts, np_parts):
        np.testing.assert_array_equal(np.asarray(part), np_part)
        assert np.asarray(part).shape == np_part.shape

    # array_split along the last axis (uneven).
    aparts = mpcf.array_split(t, 3, axis=-1)
    np_aparts = np.array_split(np.arange(12).reshape(3, 4), 3, axis=-1)
    assert len(aparts) == len(np_aparts)
    for part, np_part in zip(aparts, np_aparts):
        np.testing.assert_array_equal(np.asarray(part), np_part)
        assert np.asarray(part).shape == np_part.shape
```

</details>

### LINKED ISSUE #21: [joinsplit] split/array_split with a negative index entry raises a low-level pybind TypeError instead of

### Summary
The docstring describes numpy-like split-index semantics and the implementation faithfully matches numpy for every other index edge case I tried (out-of-range clamps to empty parts, unsorted gives empty/overlapping parts, duplicates give empty parts, leading 0 gives an empty first part). Negative indices are the one numpy behavior that instead produces a confusing TypeError, so the API is inconsistent with its own documented numpy-likeness and with its handling of all other indices.

### Minimal reproduction
```python
t = mpcf.FloatTensor(np.arange(6, dtype=np.float64))
try:
    parts = mpcf.split(t, [-2])
    print('mpcf split [-2]:', [np.asarray(p).tolist() for p in parts])
except TypeError as e:
    print('mpcf split [-2] TypeError:', str(e).splitlines()[0])
print('numpy ref split [-2]:', [p.tolist() for p in np.split(np.arange(6, dtype=np.float64), [-2])])
# -> mpcf split [-2] TypeError: split_indices(): incompatible function arguments. The following argument types are supported:
# -> numpy ref split [-2]: [[0.0, 1.0, 2.0, 3.0], [4.0, 5.0]]
# array_split(t, [-2]) raises the same TypeError.
```

### Observed
mpcf.split(t, [-2]) and mpcf.array_split(t, [-2]) raise `TypeError: split_indices(): incompatible function arguments. The following argument types are supported: ...` — a low-level pybind argument-conversion error, because the binding takes std::vector<size_t> and a negative entry cannot convert to unsigned. I independently confirmed (in /tmp, against the installed package) that every OTHER index edge case the reporter cited matches numpy exactly: out-of-range [10] -> [whole, []], unsorted [4,2] -> [(0..3),[],(2..5)], duplicate [3,3] -> [(0..2),[],(3..5)], leading-0 [0,3] -> [[], (0..2), (3..5)], and empty [] (already tested in test_tensor_split.py). Only the negative case diverges, and it diverges as a confusing TypeError rather than a meaningful result or error.

### Expected
The docstrings state indices_or_sections "gives the indices where splits occur" and the module otherwise mirrors numpy for all index edge cases, so a user reasonably expects numpy semantics: np.split(arange(6), [-2]) -> [[0,1,2,3],[4,5]] (split at axis_size+idx). At minimum, the API should raise a clear ValueError ("split indices must be non-negative") rather than a low-level pybind "incompatible function arguments" TypeError that leaks the C++ overload signature. Supporting numpy-style negative offsets (normalize idx -> axis_size + idx before dispatch) would make the documented numpy-likeness fully consistent.

### Root-cause hint
src/python/py_tensor.hpp:260 split_indices binding declares `const std::vector<size_t>& indices` (unsigned), so any negative entry fails pybind conversion; the underlying mpcf::split (include/mpcf/tensor.tpp:1000, signature in tensor.hpp:584) also uses std::vector<size_t> and has no negative-index normalization, and masspcf/tensor_create.py:split/array_split pass the raw list straight through without normalizing negatives against axis size.

In `test/python/test_bugscan_joinsplit.py` -- run with `cd test && python -m pytest python/test_bugscan_joinsplit.py`

```python
# Bug 20: split/array_split with a negative index entry raise a low-level pybind
#         TypeError instead of numpy-style from-the-end offset.
def test_negative_split_index_offsets_like_numpy():
    """split/array_split with a negative index must split numpy-style (axis_size+idx)."""
    # BUG: a negative entry in the indices list (e.g. [-2]) raises
    #      "split_indices(): incompatible function arguments" because the binding
    #      takes std::vector<size_t>; every other index edge case already matches numpy.
    # Expected: numpy semantics -- np.split(arange(6), [-2]) -> [[0,1,2,3],[4,5]]
    #      (split at axis_size + idx).
    # Observed today: low-level pybind TypeError leaking the C++ overload signature.
    arr = np.arange(6, dtype=np.float64)
    np_parts = np.split(arr, [-2])

    t = mpcf.FloatTensor(arr)
    for op in (mpcf.split, mpcf.array_split):
        parts = op(t, [-2])
        assert len(parts) == len(np_parts)
        for part, np_part in zip(parts, np_parts):
            np.testing.assert_array_equal(np.asarray(part), np_part)
            assert np.asarray(part).shape == np_part.shape
```

</details>

### LINKED ISSUE #73: [friction/joinsplit] No negative-axis support on concatenate/split/array_split forces manual axis arithmetic

### Scenario
A user used to numpy writes mpcf.concatenate(parts, axis=-1) or mpcf.split(t, n, axis=-1) on a tensor whose ndim they do not want to hardcode (e.g. a generic helper that joins/splits along 'the last axis'). stack already accepts axis=-1, so they reasonably assume the others do too.

### Current approach
```
axis=-1 raises a pybind TypeError. The user must compute the positive axis themselves: ax = axis % t.ndim (or ax = t.ndim + axis), then call mpcf.split(t, n, axis=ax). For concatenate there is no .ndim on the list, so they must do tensors[0].ndim + axis. Example workaround that runs:
  t = mpcf.FloatTensor(np.arange(12, dtype=np.float64).reshape(3,4))
  axis = -1
  ax = axis % t.ndim   # -> 1
  parts = mpcf.split(t, 2, axis=ax)
```

### Desired API
concatenate(tensors, axis=-1), split(t, sections, axis=-1), array_split(t, sections, axis=-1) should accept negative axis and resolve it like numpy / like the existing stack(axis=-1). Change the C++ bindings to take a signed axis (ptrdiff_t) and normalize, matching stack.

### Notes
stack already does exactly this (ptrdiff_t axis, normalized in tensor.tpp:978), so the fix is to bring the other three in line. This is the same root cause as the medium bug above; listing as friction because the user-facing pain is 'I can't write axis=-1'.

### LINKED ISSUE #74: [friction/joinsplit] Joining tensors of different dtype/precision (e.g. float32 + float64, or int + float) has no

### Scenario
A user has two tensors built at different precisions (one pcf32, one pcf64, or float32 vs float64, or int64 vs float64) and wants to concatenate/stack them, expecting numpy-style upcasting to a common dtype.

### Current approach
```
mpcf.concatenate([f32, f64]) raises `TypeError: concatenate(): incompatible function arguments ... Invoked with: [<Float32Tensor>, <Float64Tensor>], 0`. There is no documented common-dtype rule and the error does not say 'dtypes must match' or suggest .astype(). The user must discover .astype() on their own and pre-convert: mpcf.concatenate([f32.astype(mpcf.float64), f64]).
```

### Desired API
Either (a) raise a clear ValueError naming the mismatch, e.g. 'concatenate: all tensors must have the same dtype (got float32 and float64); cast with .astype() first', or (b) follow numpy and upcast to a common dtype automatically. At minimum the error should be actionable rather than dumping the pybind overload signature.

### Notes
Mixed families (float+pcf, pcf+pcloud) correctly cannot be joined and erroring is right -- only the message quality is the issue there. The same-family / different-precision case is where a user most plausibly expected it to just work (numpy-like upcasting).

bwehlin:
## Summary

`concatenate` / `split` / `array_split` bind an unsigned axis and unsigned split indices, so several NumPy-natural calls leaked raw pybind errors:

- **#20 / #73** a negative `axis` (`concatenate(..., axis=-1)`, `split(..., axis=-1)`) raised `incompatible function arguments`.
- **#21** a negative split-index entry (`split(t, [-2])`) raised the same pybind error.
- **#74** joining tensors of different dtype/precision (`concatenate([f32, f64])`) leaked the pybind overload dump with no clear message.

## Fix (Python wrapper, `tensor_create.py`)

- Negative `axis` now resolves from the end via the existing `_resolve_axis` helper (the same one `transpose`/`squeeze` use), for `concatenate`, `split`, and `array_split`. Out-of-range axis now raises `IndexError` (NumPy's `AxisError` subclasses it).
- Negative split-index entries offset from the end of the axis and clamp at 0, exactly matching `numpy.split` / `numpy.array_split` slice semantics; positive overflow is still clamped by the C++ layer.
- `concatenate` and `stack` validate that all inputs share a dtype and raise a clear `TypeError` otherwise (there is no implicit upcasting — cast with `astype` first), covering both precision mismatch and cross-family joins.

`stack` already accepted a negative/new axis via its `ptrdiff_t` binding and is unchanged except for the dtype check.

## Tests

`test/python/test_bugscan_joinsplit.py` extended: negative-axis concatenate/split/array_split match NumPy; out-of-range axis raises `IndexError`; negative split indices (`[-2]`, `[-7]`, `[2,-1]`, `[-3,-1]`) match `numpy.split`/`array_split` for both splitters; mismatched-precision and cross-family joins raise a clear `TypeError`. Docs (`docs/tensors.rst`) updated.

Full Python suite passes (1815 passed).

Closes #20
Closes #21
Closes #73
Closes #74


bwehlin:
## 🔍 Adversarial review

### Summary

This PR moves negative-`axis` resolution and negative split-index normalization into the Python wrapper (`stablebear/tensor_create.py`) for `concatenate`/`stack`/`split`/`array_split`, and adds an up-front dtype-equality check on the join ops. I tried hard to break it — including the cases the tests *don't* cover (extreme/non-monotonic negative indices) — and the behavior holds up. **Verdict: solid, ship-able.** Findings below are minor.

### 🟡 Edge cases to confirm (verified correct, but untested)

- **`_normalize_split_indices` with extreme / non-monotonic / clamped indices.** The tests only exercise monotonically-increasing post-resolution index lists (`[2,-1]`→`[2,6]`, `[-3,-1]`→`[4,6]`). I checked the gnarlier cases against NumPy by emulating the clamp-at-0 + the C++ boundary slicing in `sb::split` (`include/sbear/tensor.tpp:1019` — `if (sp > axis_size) sp = axis_size;`, reversed boundaries yield empty slices):
  - `[-100]`, `[-8]` (resolve below 0, clamp to 0) → matches `np.split`
  - `[-100, -1]` → matches
  - `[-1, -3]` (resolves to `[6, 4]`, **non-monotonic**, produces an empty middle part) → matches
  
  All match NumPy, so the logic is correct — but a clamp/non-monotonic case (e.g. `[-1, -3]` and `[-100]`) would be worth a one-line addition to `test_negative_split_indices_offset_like_numpy` to lock the behavior in, since it's the most divergence-prone path.

### 🟢 Nits

- **`tensors` is iterated multiple times in `concatenate`/`stack`.** `_require_same_dtype(tensors, ...)` consumes the iterable, then `[t._data for t in tensors]` consumes it again. Fine for the documented list/tuple inputs; a generator would silently break (the new dtype pass would exhaust it, leaving an empty list comp). This is **pre-existing** (the prior `if not tensors:` guard is already always-False for generators), not a regression — just noting it since the dtype check adds a second pass.
- **`_normalize_split_indices` evaluates `int(idx)` twice per element** (`int(idx)+axis_size if int(idx)<0 else int(idx)`). Cosmetic; a local `i = int(idx)` would read cleaner.

### Things I checked that are fine

- `_resolve_axis` (`_tensor_base.py:140`) uses `operator.index` and raises `IndexError`, so out-of-range positive/negative axes (and float axes) are handled cleanly for all four ops — `concatenate`'s `size_t` binding (`py_tensor.hpp:235`) never sees a negative now.
- `stack` resolves against `ndim + 1`, giving the correct NumPy `[-(ndim+1), ndim]` range; resolved axis is always non-negative so the `ptrdiff_t` binding (`py_tensor.hpp:238`) sees no double-resolution.
- The dtype check relies on `t.dtype` being present on every family — confirmed (`base_tensor.py` sets it per family; `distance_matrix.py`/`symmetric_matrix.py` expose it as a property), and dtype sentinels are hashable so `set(...)` works. Cross-family joins (e.g. float + pcf) are caught before reaching the mixed-type C++ overload.
- Integer-sections path resolves axis before dispatch, so negative axis works for `split_sections`/`array_split` too; positive index overflow is still clamped C++-side.

Nice, tightly-scoped fix with tests that actually exercise the claimed behavior against NumPy as an oracle.

---

PR #140, opened 2026-06-26
Numeric cross-dtype comparisons promote via NumPy (#49)

### LINKED ISSUE #49: [interop] Cross-dtype comparisons and array_equal between numeric tensors leak the pybind11 overload

### Summary
NumPy-semantics divergence plus a misleading/leaky error message. Comparing a float tensor to an int tensor is an extremely common operation that fails opaquely.

### Minimal reproduction
```python
f = mpcf.FloatTensor(np.array([1.0, 2.0, 3.0]))
i = mpcf.IntTensor(np.array([1, 2, 3]))
try:
    f == i
except TypeError as e:
    print("EQ:", str(e))          # leaks: __eq__(): incompatible function arguments ... masspcf._mpcf_cuda12.Float64Tensor ...
try:
    f.array_equal(i)
except TypeError as e:
    print("AE:", str(e))          # leaks: array_equal(): incompatible function arguments ...
# also leaks for !=, <, <=, >, >= and for int32-vs-int64 / float32-vs-float64
```

### Observed
FloatTensor([1.,2.,3.]) == IntTensor([1,2,3]) raises TypeError with the raw pybind11 overload table, exposing internal C++ type names: "__eq__(): incompatible function arguments. The following argument types are supported: 1. (self: masspcf._mpcf_cuda12.Float64Tensor, arg0: masspcf._mpcf_cuda12.Float64Tensor) -> masspcf._mpcf_cuda12.BoolTensor. Invoked with: <...Float64Tensor object...>, <...Int64Tensor object...>". The same leaked signature dump occurs for array_equal() and for !=, <, <=, >, >=. Scope is broader than reported: it also fires for int32-vs-int64 and float32-vs-float64 (same NumericTensor subclass, different C++ precision). It is NOT specific to float-vs-int. (Separately observed but out of scope: cross-dtype arithmetic +,-,* leak the same way, and f == 2.0 against a Python scalar wrongly returns plain False instead of elementwise.)

### Expected
A clean, user-facing error or a cast/promotion of the operands -- not a dump of the C++ overload table with internal pybind11 type names. The library already establishes a cast-to-destination-dtype convention for cross-dtype operands elsewhere (test_indexing_parity_misc.py::test_cross_dtype_tensor_rhs_casts pins that IntTensor assigned into FloatTensor casts to float, matching NumPy), so casting the comparison operands to a common dtype would be consistent with the library's own behavior. NumPy reference: float==int and int32==int64 return elementwise [True,True,True].

### Root-cause hint
masspcf/_tensor_base.py: Tensor.__eq__/__ne__/__lt__/__le__/__gt__/__ge__ (lines 580-606) and Tensor.array_equal (line 621) call self._data <op> rhs._data; the C++ pybind11 overload only accepts a same-typed argument, so a differently-typed rhs._data triggers the signature dump. NumericTensor.__eq__/array_equal (masspcf/tensor.py lines 97-105) only special-case np.ndarray (wrapping in type(self)) and do not cast/promote a cross-dtype Tensor operand, so it falls through to the leaking C++ call. Fix: in the Python wrappers, promote/cast operands to a common dtype (as setitem does) or raise a clean TypeError before reaching pybind11.

```python
# Bug 49: cross-dtype comparison / array_equal leak the pybind11 overload table.
def test_cross_dtype_comparison_no_pybind_signature_leak():
    """Cross-dtype == / array_equal must not dump the raw pybind11 overload table."""
    # BUG: float-vs-int (and int32-vs-int64, float32-vs-float64) == / array_equal
    #      raise a TypeError dumping internal C++ pybind11 type names
    # Expected: a clean user-facing result (cast/promote to a common dtype, NumPy
    #           parity: float==int -> [True,True,True]) or a clean error WITHOUT
    #           leaking C++ overload signatures / internal type names.
    # Observed today: TypeError containing "incompatible function arguments" and
    #           internal names like "masspcf._mpcf_cuda12.Float64Tensor".
    f = mpcf.FloatTensor(np.array([1.0, 2.0, 3.0]))
    i = mpcf.IntTensor(np.array([1, 2, 3]))

    try:
        r = f == i
    except TypeError as e:
        msg = str(e)
        assert "incompatible function arguments" not in msg, (
            f"leaked pybind11 overload table: {msg}")
        assert "Tensor" not in msg or "masspcf._mpcf" not in msg, (
            f"leaked internal C++ type names: {msg}")
    else:
        assert isinstance(r, mpcf.BoolTensor), (
            f"expected BoolTensor, got {type(r).__name__}={r!r}")
        np.testing.assert_array_equal(
            np.asarray(r), np.array([1.0, 2.0, 3.0]) == np.array([1, 2, 3]))

    try:
        ae = f.array_equal(i)
    except TypeError as e:
        msg = str(e)
        assert "incompatible function arguments" not in msg, (
            f"leaked pybind11 overload table from array_equal: {msg}")
    else:
        assert ae == np.array_equal(
            np.array([1.0, 2.0, 3.0]), np.array([1, 2, 3]))
```

</details>

bwehlin:
## Summary

Comparing two numeric tensors of different dtype leaked the raw pybind overload table, because the C++ comparison bindings only accept a same-typed right-hand side:

```python
sb.FloatTensor([1.,2.,3.]) == sb.IntTensor([1,2,3])   # TypeError: __eq__(): incompatible function arguments ...
```

This affected `==`, `!=`, the ordering operators, and `array_equal`. (The scalar case `t == 2.0` was already fixed in #122.)

## Fix

`NumericTensor._compare` and `NumericTensor.array_equal` now detect a `NumericTensor` RHS of a different dtype and evaluate through NumPy (`op(np.asarray(self), np.asarray(rhs))`), returning a `BoolTensor`. NumPy `result_type` promotion gives by-value semantics — `IntTensor([1]) == FloatTensor([1.5])` is `False`, not a silent down-cast to equal. Same-dtype comparisons keep the existing C++ fast path, and the scalar/ndarray path is unchanged.

## Tests

New `test/python/test_bugscan_cross_dtype_compare.py` (31 cases): all six comparison operators across float32/float64, int32/int64, int/float, and uint64/int64 match NumPy; `array_equal` matches NumPy across dtype mixes; by-value correctness (`1 == 1.5` is `False`); no pybind leak; same-dtype path unaffected; cross-dtype `t[t == u]` mask-select works.

Full Python suite passes (1830 passed).

Closes #49


bwehlin:
## 🔍 Adversarial review

### Summary
The PR fixes cross-dtype numeric tensor comparisons (`==`, `!=`, `<`, `<=`, `>`, `>=`, `array_equal`) by detecting a differently-typed `NumericTensor` RHS in `NumericTensor._compare` / `NumericTensor.array_equal` and evaluating through NumPy, returning a `BoolTensor`. The fix is small, correct in approach, and the dtype-identity check is sound (dtype sentinels are singletons in `typing.py`, so `rhs.dtype is not self.dtype` is the right test). Verdict: **solid and mergeable**; the findings below are edge cases and nits, not blockers.

### Findings

#### 🟡 Edge cases to confirm

1. **Broadcasting/shape-mismatch behavior now diverges between the same-dtype and cross-dtype paths** (`stablebear/base_tensor.py`, new `_compare`). The cross-dtype path runs `op(np.asarray(self), np.asarray(rhs))`, which inherits full NumPy broadcasting (e.g. `FloatTensor([1,2,3]) == IntTensor([[1],[2]])` would produce a `(2,3)` result, and shape-incompatible operands raise NumPy's `ValueError`). The same-dtype path goes through the C++ `op(self._data, rhs._data)` binding, which may have different (stricter or non-broadcasting) shape semantics. So whether two numeric tensors broadcast on comparison now silently depends on whether their dtypes happen to match. Worth confirming the same-dtype C++ path broadcasts identically (or documenting that comparisons require equal shapes).

2. **`uint64` vs signed `int64` promotes to `float64`** — the explicitly-tested case. NumPy has no common integer type for `uint64`+`int64`, so it promotes both to `float64`, which loses precision above 2^53. The test only uses `[1,2,3]`, so it cannot detect a wrong answer for large magnitudes (e.g. `uint64(2**63+1) == int64(...)`). This matches NumPy's (lossy) behavior, so it's arguably "correct by spec," but the by-value-correctness guarantee in the PR description quietly does not hold for large 64-bit integers. A one-line comment or test acknowledging the float64 promotion limit would be honest.

#### 🟢 Nits

3. **Tautological assertions in the parametrized tests.** `test_cross_dtype_comparison_matches_numpy` asserts `result == np_op(np.asarray(lhs), np.asarray(rhs))`, but the implementation *is* `op(np.asarray(self), np.asarray(rhs))` — so this can essentially never fail; it confirms the code delegates to NumPy rather than that the delegation is the *desired* answer. Same for the second assertion in `test_cross_dtype_array_equal_matches_numpy`. The genuinely valuable tests here are the hardcoded-expectation ones (`1 == 1.5` is `False`, the `no_pybind_overload_leak` guard, and the `t[t == u]` mask-select idiom). Consider adding a couple more hardcoded-expected-value cases (especially an ordering op like `<` across int/float with a non-trivial expected mask) rather than relying on the NumPy-mirrors-NumPy comparisons.

4. **No coverage for a non-numeric Tensor RHS.** A `BoolTensor` (or other non-`NumericTensor` `Tensor`) RHS of differing dtype still falls through to `super()._compare`, i.e. the C++ binding, and can still leak the pybind "incompatible function arguments" error — the exact failure mode this PR fixes, just for `bool`-vs-numeric. Out of scope for issue #49 (which is numeric-vs-numeric), but it leaves a closely-adjacent leak unaddressed; a test asserting the current behavior (or a follow-up note) would make the boundary explicit.

5. **Redundant outer `np.asarray`** in `BoolTensor(np.asarray(op(np.asarray(self), np.asarray(rhs))))` — `op(...)` already returns an ndarray. Harmless, just noise.

Nothing here rises to a likely correctness bug given the stated scope; the core fix does what the description claims.

---

Issue #63, opened 2026-06-07
[friction/arithmetic] Mixed-precision tensor-tensor arithmetic raises a raw pybind error instead of promoting


bwehlin:
_Usability friction found by the masspcf API scan (2026-06-07). Not a correctness bug; see `friction-triage.md`._  **Priority: medium.**

### Scenario
A user adds a float32 tensor to a float64 tensor (or int32 to int64, or int to float), expecting NumPy-style dtype promotion.

### Current approach
```
a32 = mpcf.FloatTensor(np.array([1.,2.,3.], dtype=np.float32))
a64 = mpcf.FloatTensor(np.array([10.,20.,30.], dtype=np.float64))
a32 + a64  # TypeError: __add__(): incompatible function arguments ... (Float32Tensor vs Float64Tensor)
# Workaround:
a32.astype(mpcf.float64) + a64
```

### Desired API
Numeric tensor binary ops should promote operand dtypes NumPy-style (float32+float64 -> float64, int32+float64 -> float64, int32+int64 -> int64), or at minimum raise a clean, documented TypeError naming the two dtypes instead of dumping the pybind overload signatures.

### Notes
The ArithmeticTensorMixin docstring advertises NumPy-style broadcasting but is silent on dtype promotion; users will reasonably expect promotion since the library positions itself as numpy-like.

Issue #62, opened 2026-06-07
[friction/arithmetic] Tensor + numpy.ndarray fails even though numpy.ndarray + tensor works (operator asymmetry)


bwehlin:
_Usability friction found by the masspcf API scan (2026-06-07). Not a correctness bug; see `friction-triage.md`._  **Priority: high.**

### Scenario
A user has a FloatTensor f and wants to add a numpy array elementwise: f + np.array([...]). The reverse (np.array + f) works, so the asymmetry is surprising.

### Current approach
```
f = mpcf.FloatTensor(np.array([1.,2.,3.]))
arr = np.array([10.,20.,30.])
# np.array + f  -> works (uses __radd__): [11,22,33]
# f + np.array  -> TypeError: __add__(): incompatible function arguments ... Invoked with: <Float64Tensor>, array([10.,20.,30.])
# Workaround: wrap the array first
f + mpcf.FloatTensor(arr)   # or:  arr + f
```

### Desired API
__add__/__sub__/__mul__/__truediv__ (and the in-place variants) on numeric tensors should accept a numpy.ndarray RHS and wrap it (the same way __setitem__ already does via type(self)(val)), so f + arr behaves like arr + f. The raw pybind TypeError dump should be replaced with a clean message if the type is truly unsupported.

### Notes
__array_ufunc__=None correctly routes np.array + f to __radd__, but the forward direction has no ndarray handling, so the two directions disagree. __setitem__ already demonstrates the wrap-as-tensor pattern.

PR #141, opened 2026-06-26
Clear error for mismatched-dtype tensor arithmetic (#64)

### LINKED ISSUE #64: [friction/arithmetic] PcfTensor + PcfTensor dtype mismatch gives an ugly pybind error, unlike the clean Pcf + Pcf

### Scenario
A user adds a pcf32 tensor to a pcf64 tensor (or pcf32 to pcf32i). Pcf+Pcf scalar mismatch gives a clean 'Mismatched PCF types' TypeError (tested in test_pcf_add.py), but the tensor-level op does not.

### Current approach
```
X32 = mpcf.zeros((2,), dtype=mpcf.pcf32)
X64 = mpcf.zeros((2,), dtype=mpcf.pcf64)
X32 + X64  # TypeError: __add__(): incompatible function arguments ... (Pcf32Tensor vs Pcf64Tensor)
# vs scalar:
# mpcf.Pcf(f32) + mpcf.Pcf(f64) -> TypeError('Mismatched PCF types')   # clean
```

### Desired API
PcfTensor/IntPcfTensor arithmetic should raise the same clean 'Mismatched PCF types' (or 'All inputs must have the same dtype') TypeError that _resolve_pcf_inputs and Pcf+Pcf already produce, rather than the raw pybind11 overload dump.

### Notes
Inconsistent with the scalar Pcf path and with _resolve_pcf_inputs' nice message; the ArithmeticTensorMixin operators bypass the dtype-checking helpers entirely.

bwehlin:
## Summary

Tensor-tensor arithmetic between different dtypes leaked the raw pybind overload dump:

```python
sb.zeros((2,), dtype=sb.pcf32) + sb.zeros((2,), dtype=sb.pcf64)
# TypeError: __add__(): incompatible function arguments ...
```

unlike the clean `Pcf + Pcf` mismatch error (`Mismatched PCF types`).

## Fix

`ArithmeticTensorMixin._decay_operand` now detects a `Tensor` right-hand side whose dtype differs from `self` and raises a clear `TypeError` naming both dtypes (`All inputs must have the same dtype (got pcf32, pcf64).`). This covers `+ - * /` and their in-place forms for every tensor family. There is no implicit promotion yet (tracked separately in #63); same-family precision mixing simply errors cleanly for now.

Preserved paths: a non-`Tensor` element such as a scalar `Pcf` (which the C++ operator broadcasts across the tensor), ndarray/list/tuple right-hand sides (#62), and same-dtype arithmetic are all unchanged — keyed on `isinstance(val, Tensor)` so the scalar `Pcf` (which has `_data` but no `.dtype`) is not caught.

## Tests

`TestMixedDtypeArithmeticRaisesClean` in `test/python/test_tensor_arithmetic.py`: `pcf32`/`pcf64` mismatch across all four operators, pcf float-vs-int mismatch, numeric `float32`/`float64` mismatch, in-place mismatch, the message is not the pybind dump, and same-dtype arithmetic + scalar-`Pcf` RHS still work.

Full Python suite passes (1809 passed).

Closes #64


bwehlin:
## 🔍 Adversarial review

### Summary
The PR makes mismatched-dtype tensor-tensor arithmetic (`pcf32 + pcf64`, `float32 + float64`, in-place forms, etc.) raise a clear `TypeError` naming both dtypes instead of leaking the raw pybind "incompatible function arguments" dump, by adding an `isinstance(val, Tensor)` dtype check to `ArithmeticTensorMixin._decay_operand`. I tried hard to break it and it holds up: the change is correct, the premise is accurate, and same-dtype / scalar-`Pcf` / array paths are genuinely preserved. **Verdict: solid, mergeable.** A couple of minor nits below.

I verified the core claims against the real code:
- **Premise is correct.** `src/python/py_tensor.hpp` only binds `__add__`/`__sub__`/`__mul__`/`__truediv__` for `(const TTensor& self, const TTensor& rhs)` of the *same* `TTensor` type (lines 281, 289, 297, 305) plus scalar overloads (line 368+). There is no cross-precision tensor overload, so `cpp.Float32Tensor + cpp.Float64Tensor` really does fall through to the pybind dump — the fix's rationale is accurate.
- **Identity comparison is safe.** dtypes are singletons (`stablebear/typing.py`), so `getattr(val, "dtype", None) is not getattr(self, "dtype", None)` is the correct comparison.
- **Scalar-`Pcf` path preserved.** `Pcf` (`stablebear/functional/pcf.py`) is a standalone class with `_data` but no `.dtype`, and is not a `Tensor`, so `isinstance(val, Tensor)` correctly skips it and it falls to the `hasattr(val, "_data")` branch — matching the test `test_scalar_pcf_rhs_still_broadcasts`.
- **Same-dtype is byte-for-byte unchanged.** The new same-dtype branch returns `val._data`, identical to the old `hasattr` path.
- **No reflected-op bypass.** The operators raise rather than return `NotImplemented`, so `a + b` never falls back to `b.__radd__(a)` to evade the check. I also confirmed no existing test relied on the old behavior or on cross-precision arithmetic working.

### 🟡 Edge cases to confirm
- **Reflected-op message operand order** (`stablebear/_tensor_base.py`, `_decay_operand`): in `__rsub__`/`__rtruediv__` the message is built as `(got {self.dtype.name}, {val.dtype.name})`, where `val` is the *left* operand. So for a reflected mismatch the two dtypes appear in reversed source order. It's only reachable with a mismatched-dtype `Tensor` on the left whose type lacks the operator (e.g. `PointCloudTensor - FloatTensor` → `FloatTensor.__rsub__`), which is niche, but the message would read backwards relative to how the user wrote it. Both dtypes are still named, so this is cosmetic.

### 🟢 Nits
- **Test coverage of the new branch could be a touch broader.** The tests exercise the same-family mismatch well, but not (a) a cross-*family* pair such as `PointCloudTensor + FloatTensor`, which now also routes through the new branch and produces the clean error via `__radd__`, nor (b) the reflected/division mismatch path. Adding one such case would lock in the broader behavior the diff actually enables.
- **Defensive f-string assumption.** The error message does `self.dtype.name` / `val.dtype.name` unconditionally after the `getattr(..., None)` guard. If a `Tensor` instance ever lacked a `.dtype` attribute, the error message itself would raise `AttributeError`. Every real `Tensor` subclass sets `self.dtype` in `__init__` (confirmed for Float/Int/Pcf/IntPcf/Bool), so this is purely theoretical — noting it only because the guard uses `getattr(..., None)` while the message assumes the attribute exists.

Nothing blocking. Nice that it mirrors the existing `Mismatched PCF types` UX and the CHANGELOG entry is accurate.

---

PR #142, opened 2026-06-26
NumPy-style float assignment and exponents on IntTensor (#7, #65)

### LINKED ISSUE #7: [setitem] Python-float scalar/slice assignment into IntTensor raises a raw pybind11 TypeError dump instead of

### Summary
Inconsistent and misleading behavior: a value type the library explicitly declares valid for setitem (float) passes validation but then fails inside C++ with a raw, non-Pythonic error, and the failure mode differs from the equivalent tensor-RHS path which truncates successfully. This is a wrong-kind/misleading-error issue and a self-inconsistency in the documented setitem contract.

### Minimal reproduction
```python
import numpy as np
from masspcf.tensor import IntTensor, FloatTensor

# Cross-dtype tensor RHS truncates cleanly (works)
a = IntTensor(np.array([0, 0, 0], dtype=np.int64))
a[:] = FloatTensor(np.array([7.9, 8.9, 9.9], dtype=np.float32))
print('tensor RHS:', np.asarray(a).tolist())          # [7, 8, 9]

# float is declared a valid setitem dtype for IntTensor
b = IntTensor(np.array([1, 2, 3], dtype=np.int64))
print('float valid:', float in b._get_valid_setitem_dtypes())  # True

# scalar float element assign -> raw pybind11 TypeError
try:
    b[0] = 7.9
    print('element:', np.asarray(b).tolist())
except TypeError as e:
    print('element ERR:', str(e)[:55])   # _set_element(): incompatible function arguments...

# scalar float slice broadcast-fill -> raw pybind11 constructor TypeError
c = IntTensor(np.array([1, 2, 3, 4], dtype=np.int64))
try:
    c[0:2] = 7.9
    print('slice:', np.asarray(c).tolist())
except TypeError as e:
    print('slice ERR:', str(e)[:55])     # __init__(): incompatible constructor arguments...

# NumPy reference truncates
npa = np.array([1, 2, 3]); npa[0] = 7.9
print('numpy ref:', npa.tolist())        # [7, 2, 3]
# (run from /tmp; exit code 0, no crash)
```

### Observed
Assigning a Python float to an IntTensor element or slice raises a raw pybind11 overload-resolution / constructor dump. b[0] = 7.9 -> "TypeError: _set_element(): incompatible function arguments. The following argument types are supported..."; c[0:2] = 7.9 -> "TypeError: __init__(): incompatible constructor arguments...". This happens despite float being in IntTensor._get_valid_setitem_dtypes() (so _validate_setitem_dtype passes it through), and despite the cross-dtype tensor-RHS path (a[:] = FloatTensor([7.9, 8.9, 9.9])) truncating cleanly to [7, 8, 9]. Reproduced exactly from /tmp against the installed package at /opt/venv/lib/python3.10/site-packages/masspcf (verified byte-identical to the source tree). No crash; process exit code 0.

### Expected
Consistent behavior: either truncate the Python float to int (NumPy parity -- np.array([1,2,3]); a[0]=7.9 -> [7,2,3] -- and matching the library's own cross-dtype tensor-RHS truncation), or raise the library's own clean Pythonic TypeError from _validate_setitem_dtype. The same logical operation should not succeed via tensor RHS while failing with a low-level pybind signature dump via scalar/slice RHS.

### Root-cause hint
masspcf/tensor.py:NumericTensor._decay_value (line 82-83) is a no-op pass-through (return val) shared by both FloatTensor and IntTensor; IntTensor never overrides it. With float listed in NumericTensor._get_valid_setitem_dtypes (line 79-80), validation accepts the float, then _basic_setitem in masspcf/_tensor_base.py passes the raw float to the C++ Int tensor _set_element (element path, line 552) / type(self._data)(Shape, val) constructor (slice broadcast-fill path, line 571), which require ints. Fix: have IntTensor override _decay_value to coerce float/np.floating to int (truncation, matching _coerce_rhs/astype on the tensor-RHS path), or otherwise route scalar float assignment through the same casting logic.

In `test/python/test_bugscan_setitem.py` -- run with `cd test && python -m pytest python/test_bugscan_setitem.py`

```python
def test_float_scalar_into_inttensor_no_raw_pybind_error():
    """Python-float scalar/slice assignment into IntTensor must not emit a raw pybind dump."""
    # BUG: Python-float scalar/slice assignment into IntTensor raises a raw
    #      pybind11 TypeError dump instead of truncating (NumPy parity, matching
    #      the cross-dtype tensor-RHS path) or raising a clean library error.
    # Expected: either truncate to int (NumPy parity -- np.array([1,2,3]);
    #           a[0]=7.9 -> [7,2,3]) like the working tensor-RHS path, OR raise
    #           the library's own clean Pythonic TypeError from
    #           _validate_setitem_dtype. The raw pybind11 overload/constructor
    #           "incompatible ... arguments" dump is never acceptable, and float
    #           is explicitly declared a valid setitem dtype for IntTensor.
    # Observed today: b[0] = 7.9 -> "TypeError: _set_element(): incompatible
    #                 function arguments ..."; c[0:2] = 7.9 -> "TypeError:
    #                 __init__(): incompatible constructor arguments ...".

    # Sanity: the equivalent cross-dtype tensor-RHS path already truncates.
    ref = IntTensor(np.array([0, 0, 0], dtype=np.int64))
    ref[:] = FloatTensor(np.array([7.9, 8.9, 9.9], dtype=np.float32))
    assert np.asarray(ref).tolist() == [7, 8, 9]

    # Sanity: float is declared a valid setitem dtype for IntTensor.
    probe = IntTensor(np.array([1, 2, 3], dtype=np.int64))
    assert float in probe._get_valid_setitem_dtypes()

    # Scalar element assignment: must not leak a raw pybind11 signature dump.
    b = IntTensor(np.array([1, 2, 3], dtype=np.int64))
    try:
        b[0] = 7.9
    except TypeError as exc:
        # Acceptable only if it's a clean library error, not a pybind dump.
        msg = str(exc)
        assert "incompatible function arguments" not in msg
        assert "incompatible constructor arguments" not in msg
    else:
        # If it succeeded, it must have truncated to NumPy parity.
        assert np.asarray(b).tolist() == [7, 2, 3]

    # Slice broadcast-fill assignment: same contract.
    c = IntTensor(np.array([1, 2, 3, 4], dtype=np.int64))
    try:
        c[0:2] = 7.9
    except TypeError as exc:
        msg = str(exc)
        assert "incompatible function arguments" not in msg
        assert "incompatible constructor arguments" not in msg
    else:
        assert np.asarray(c).tolist() == [7, 7, 3, 4]
```

</details>

### LINKED ISSUE #65: [friction/arithmetic] Fractional exponent on an IntTensor raises a raw pybind error instead of promoting to float

### Scenario
A user wants the elementwise square root of an integer tensor: int_tensor ** 0.5.

### Current approach
```
t = mpcf.IntTensor(np.array([4,9,16], dtype=np.int64))
t ** 0.5  # TypeError: __pow__(): incompatible function arguments (Int64Tensor only accepts SupportsInt)
# Workaround:
t.astype(mpcf.float64) ** 0.5
```

### Desired API
IntTensor.__pow__ with a non-integer exponent should promote to float64 and return a FloatTensor (matching NumPy: np.array([4,9,16])**0.5 -> float64 [2,3,4]), instead of failing with a raw pybind error.

### Notes
Relatedly, IntTensor ** -1 silently returns 0 (computed as float 0.5 truncated to int) where NumPy raises ValueError 'Integers to negative integer powers are not allowed' -- a quiet wrong result rather than a clean error.

bwehlin:
## Summary

Two `IntTensor` operations leaked raw pybind errors or silently produced wrong results instead of following NumPy:

- **#7** `t[i] = 7.9` / `t[a:b] = 7.9` raised `_set_element(): incompatible function arguments` / `__init__(): incompatible constructor arguments`.
- **#65** `int_tensor ** 0.5` raised `__pow__(): incompatible function arguments`, and `int_tensor ** -1` silently returned `0` (float `pow` cast back to int).

## Fix (Python wrapper on `IntTensor`)

- `_decay_value` truncates a Python/NumPy float toward zero when it reaches the scalar/slice fill path, so `t[0] = 7.9` stores `7` and `t[0] = -7.9` stores `-7`, matching NumPy integer assignment. (`np.float64` is covered; `np.float32` remains rejected at validation, a separate pre-existing gap.)
- `__pow__` / `__ipow__`: a fractional (or whole-float) exponent promotes to a `float64` `FloatTensor` (`t ** 0.5`, `t ** 2.0`); a negative integer exponent raises `ValueError("Integers to negative integer powers are not allowed.")`; a non-negative integer exponent keeps the integer path. In-place power with a fractional exponent raises `TypeError` (the result can't be stored in place), consistent with the existing `__itruediv__` behaviour.

## Tests

`test/python/test_int_tensor.py` gains `TestIntTensorFloatAssignTruncates` (scalar/slice/negative/`np.float64` truncation across int32/int64) and `TestIntTensorPow` (fractional → float64, whole-float → float64, non-negative int → int, in-place fractional raises, in-place int works, across all four int dtypes) plus a negative-exponent `ValueError` test.

Full Python suite passes (1829 passed).

Closes #7
Closes #65


bwehlin:
## 🔍 Adversarial review

### Summary
This PR makes float scalar/slice assignment into an `IntTensor` truncate toward zero (NumPy-style) and routes `IntTensor` exponentiation correctly: fractional/whole-float exponents promote to a `float64` `FloatTensor`, negative-integer exponents raise, and non-negative-integer exponents keep the int path. I traced both fixes against the real code (`base_tensor.py`, the `_setitem` path in `_tensor_base.py`, and the C++ `__pow__` binding in `py_tensor.hpp`/`math.hpp`) and against NumPy reference behavior. **The fix is correct and the core claims hold up.** I found no likely bugs — only a few edge-case mismatches and nits worth confirming.

I verified the headline mechanics:
- `_decay_value` uses `int(val)`, which truncates toward zero, so `7.9 → 7` and `-7.9 → -7`, matching NumPy. ✅
- The C++ pow binding (`py_tensor.hpp:437`) takes `Tv = scalar_of_t<T>` (the *integer* element type for an int tensor), which is exactly why pre-fix `t ** 0.5` raised "incompatible function arguments" and `t ** -1` did `static_cast<int>(std::pow(v,-1)) == 0`. The Python override now intercepts both before they reach C++. ✅
- The `ValueError` message is byte-for-byte NumPy's: `"Integers to negative integer powers are not allowed."` ✅
- MRO confirms `super().__pow__` / `super().__ipow__` resolve to `ArithmeticTensorMixin`, and `__ipow__` returns `self`. ✅

### 🟡 Edge cases to confirm

- **Negative integer exponent on an *unsigned* tensor diverges from NumPy.** `base_tensor.py` `__pow__`/`__ipow__` raise `ValueError("Integers to negative integer powers are not allowed.")` uniformly for all four int dtypes. But NumPy raises a different error for unsigned types:
  ```
  np.array([2,3], dtype=np.uint32) ** -1
  # OverflowError: Python integer -1 out of bounds for uint32
  ```
  So the PR's "matching NumPy" is only accurate for `int32`/`int64`. The behavior is arguably *more* sensible than NumPy's, but the description overstates parity, and there is **no test** for a negative exponent on `uint32`/`uint64` — `test_negative_integer_exponent_raises` is parametrized only over int32/int64. Worth either a test (asserting whatever you intend) or a note.

- **Negative-float assignment into an unsigned tensor is untested and likely to error.** `TestUintFloatAssignTruncates` deliberately uses only *positive* floats. `t[0] = -7.9` on a `uint32` tensor would `_decay_value` to the Python int `-7`, which the C++ uint fill/`_set_element` will reject (out of range). That may be acceptable, but the behavior is unspecified and unexercised.

- **`np.float32` scalar assignment remains rejected while `np.float64` is accepted.** This is the asymmetry the PR already flags: `np.float64` subclasses Python `float` (so it passes `_get_valid_setitem_dtypes` → `[NumericTensor, float, int, np.ndarray]`), but `np.float32` does not, so `t[0] = np.float32(5.7)` still raises a `TypeError` from `_validate_setitem_dtype`. Fine to defer, but there's no test pinning this rejection, so the gap could silently change later.

- **NaN/inf float assignment.** `t[0] = float('nan')` → `int(nan)` raises `ValueError`; `float('inf')` → `OverflowError`. NumPy also errors here, so this is roughly consistent — just untested.

### 🟢 Nits

- `TestIntTensorFloatAssignTruncates.test_negative_float_truncates_toward_zero` guards with `if sb_dtype in (sb.int32, sb.int64):`, but the class is parametrized **only** over int32/int64, so the condition is always true — dead branch. (If the intent was to extend the class to uint later, the guard would silently skip the assertion instead of testing the uint case.)

- The in-place fractional path raises plain `TypeError`; NumPy raises `UFuncTypeError` (a `TypeError` subclass). `pytest.raises(TypeError)` matches and the semantics align, so this is fine — noting only that it's a deliberate, reasonable divergence consistent with the existing `__itruediv__`.

- Pre-existing (not this PR): integer pow goes through `static_cast<T>(std::pow(double,double))`, so large `int64`/`uint64` results past 2^53 (e.g. `IntTensor([10], int64) ** 18`) can be imprecise. Out of scope here, but the new int-pow tests use only tiny bases, so they wouldn't catch it.

### Verdict
Solid, well-targeted fix with genuinely good test coverage for the signed-int happy paths. The only substantive gap is unsigned-dtype edge behavior (negative exponents and negative-float assignment) being both undocumented-as-divergent and untested. No blocking issues.

---

PR #143, opened 2026-06-26
Give BoolTensor a readable repr/str (#96)

### LINKED ISSUE #96: [friction/interop] BoolTensor has no __repr__/__str__, so it prints as <masspcf.tensor.BoolTensor object at 0x...>

### Scenario
A user runs a comparison (mask = a < b) and prints/inspects the result -- BoolTensor is the single most common output of the comparison surface.

### Current approach
```
repr(mpcf.BoolTensor(np.array([True,False]))) -> '<masspcf.tensor.BoolTensor object at 0x...>'. To see contents you must np.asarray(mask). FloatTensor and IntTensor DO override __repr__/__str__ (they show array(...)); BoolTensor does not, despite having __array__.
```

### Desired API
BoolTensor.__repr__/__str__ that delegate to np.asarray(self).__repr__()/__str__() (mirroring FloatTensor/IntTensor), so a BoolTensor prints like array([ True, False]).

### Notes
Pure inconsistency: the same one-line override exists on the sibling numeric tensors.

bwehlin:
## Summary

`BoolTensor` (e.g. the result of `t > 3`) inherited `object.__repr__`, so it printed as `<stablebear.base_tensor.BoolTensor object at 0x...>` in the REPL and notebooks — the lone outlier among the array-like tensors (`FloatTensor`/`IntTensor` and the matrix/barcode wrappers all define `__repr__`).

## Fix

Add `__repr__`/`__str__` to `BoolTensor` that delegate to `np.asarray(self)`, mirroring the numeric tensors.

## Tests

`test/python/test_tensor_comparison.py`: `repr`/`str` show the boolean values (not the object address) and match `np.asarray`, including a `BoolTensor` produced by a comparison.

Closes #96


bwehlin:
## 🔍 Adversarial review

### Summary
The PR adds `__repr__`/`__str__` to `BoolTensor` (in `stablebear/base_tensor.py`), both delegating to `np.asarray(self).__repr__()/.__str__()`, plus three tests in `test/python/test_tensor_comparison.py` and a CHANGELOG entry. I tried to break it and could not find a real bug — the fix is correct, minimal, and consistent with the existing pattern. Verdict: **solid, ship it** (a couple of optional follow-ups below).

I verified the premise of the fix against the real code:
- `BoolTensor` extends `Tensor` directly (`base_tensor.py:388`), not `NumericTensor`.
- The base `Tensor` (ABC, `_tensor_base.py:150`) defines no `__repr__`/`__str__`, so `BoolTensor` genuinely inherited `object.__repr__` → the `<...object at 0x...>` bug is real.
- `BoolTensor` already overrides `__array__` (`base_tensor.py:429`), so the `np.asarray(self)` delegation has a valid backing implementation.
- The new methods exactly mirror `FloatTensor` (`base_tensor.py:137-141`) and `IntTensor` (`204-208`), so behavior is consistent across the array-like tensors.
- `FloatTensor(...) > 2.0` does return a `BoolTensor` (`_tensor_base.py:649-650` → `_compare` → `BoolTensor(...)`), so `test_bool_tensor_from_comparison_repr`'s `isinstance` assertion is valid, and `BoolTensor` is exported (`stablebear/__init__.py:16`).

### Findings

#### 🟢 Nits / optional

1. **Test `repr(t) == repr(np.asarray(t))` is partly tautological.** Since the production `__repr__` *is literally* `np.asarray(self).__repr__()`, that equality assertion re-implements the code under test and cannot catch a wrong-but-consistent implementation. It's not harmful — the regression itself is caught by the `"object at 0x" not in repr(t)` check (`test_bool_tensor_repr_shows_values`), which is the load-bearing assertion. Just be aware the equality check adds little signal.

2. **0-d / scalar `BoolTensor` is untested.** All three tests use 1-d tensors. A scalar comparison (e.g. `sb.FloatTensor(5.0) > 2.0`, or indexing that yields a 0-d `BoolTensor`) would render via `repr(np.asarray(...))` as `array(True)`. I reasoned through `__array__` (`np.array(data)` on a 0-d) and it should work fine, so this is low-risk — but given that 0-d tensors are a recurring sharp edge in this codebase (see the sibling `#133` / `#39` CHANGELOG entries), a one-line scalar `repr`/`str` case would be cheap insurance.

3. **`__str__` from a comparison is untested.** `test_bool_tensor_from_comparison_repr` exercises `repr` on a comparison result but not `str`; `test_bool_tensor_str_shows_values` only uses a literally-constructed tensor. Minor, since both delegate identically.

No 🔴 likely bugs and no 🟡 edge cases that I'd block on. The repr/str output looking identical to a plain NumPy array (`array([...])`) rather than something `BoolTensor`-branded is a pre-existing, intentional convention shared with `FloatTensor`/`IntTensor`, so not a defect of this PR.

---

PR #144, opened 2026-06-26
Pcf ergonomics: repr, dtype, array accessors, from_arrays (#66, #67, #68, #69)

### LINKED ISSUE #66: [friction/pcf] No constructor from separate times and values arrays

### Scenario
A user has a 1-D array of breakpoint times and a 1-D array of values (the natural way they computed them) and wants a Pcf. The only constructor accepts a single (n,2) array (or a Pcf, or a list).

### Current approach
```
times = np.array([0.0,1.0,3.0]); values = np.array([1.0,2.0,0.5]); f = mpcf.Pcf(np.column_stack([times, values]))  # must manually stack into (n,2)
```

### Desired API
mpcf.Pcf(times, values, dtype=...) or a classmethod mpcf.Pcf.from_arrays(times, values, dtype=...) that pairs two equal-length 1-D arrays.

### Notes
column_stack also forces a dtype promotion decision the user did not ask for (e.g. int times + float values -> float).

### LINKED ISSUE #67: [friction/pcf] No accessors for the time/value arrays or domain bounds

### Scenario
A user wants just the breakpoint times, just the values, or the support end (last breakpoint time) of a Pcf for plotting or analysis.

### Current approach
```
f.to_numpy()[:,0]  # times
f.to_numpy()[:,1]  # values
f.to_numpy()[-1,0] # last breakpoint / support start of the final interval
# no f.times, f.values, f.breakpoints, or f.support
```

### Desired API
Properties f.times, f.values (1-D arrays), and optionally f.breakpoints (the (n,2) view) plus f.t_min/f.t_max for the domain; matches how numpy-like libraries expose structure.

### Notes
to_numpy() is read-only, so slicing it is safe but verbose; the abstraction leaks the (n,2) layout.

### LINKED ISSUE #68: [friction/pcf] Pcf has no .dtype property (only .ttype/.vtype), inconsistent with tensors

### Scenario
A user who works with masspcf tensors (which expose .dtype returning pcf32/pcf64/etc.) expects the same on a single Pcf, e.g. to branch on dtype or to pass it to astype.

### Current approach
```
f.vtype  # returns float32/float64/int32/int64 scalar dtype, NOT pcf32/pcf64
# there is no f.dtype; to get the pcf-level dtype you must map vtype yourself
```

### Desired API
Pcf.dtype property returning the pcf-level dtype (pcf32/pcf64/pcf32i/pcf64i), mirroring tensor.dtype and matching what astype() consumes.

### Notes
vtype/ttype expose float32 etc., but astype() and zeros() take pcf32 etc.; the mismatch forces a manual translation.

### LINKED ISSUE #69: [friction/pcf] Pcf has no __repr__, so it shows the default object address in REPL/notebooks

### Scenario
A user inspects a Pcf interactively (just typing the variable, or in a list/notebook output).

### Current approach
```
repr(f)  # '<masspcf.functional.pcf.Pcf object at 0x...>'  -- str(f) is nice ('<PCF size=3, dtype=float64>') but repr is what the REPL/notebook shows
```

### Desired API
Define __repr__ on Pcf (e.g. return the same string as __str__, or a more informative one). Containers of Pcf and bare REPL evaluation currently print useless object addresses.

### Notes
__str__ is implemented well; only __repr__ is missing.

bwehlin:
Adds small additive, read-only ergonomics to the `Pcf` class:

- `__repr__` now returns the `__str__` summary so REPL/notebooks show the PCF summary instead of an object address. (`__str__` is unchanged.)
- A read-only `.dtype` property mapping the scalar `vtype` to the PCF-level dtype (`pcf32`/`pcf64`/`pcf32i`/`pcf64i`).
- `.times`, `.values`, `.breakpoints` array accessors and `.t_min` / `.t_max` domain bounds derived from `to_numpy()` (`t_max` is the last breakpoint time, not infinity).
- A `Pcf.from_arrays(times, values, dtype=None)` classmethod that validates 1-D, equal-length inputs, column-stacks them, and delegates to `Pcf()` so the existing t0=0 / strictly-increasing validation still runs.

Includes tests in `test/python/test_pcf.py`, prose docs in `docs/concepts.rst`, and a CHANGELOG entry. Full Python suite: 1813 passed.

Closes #66
Closes #67
Closes #68
Closes #69


bwehlin:
## 🔍 Adversarial review

**Summary.** This PR adds small, additive, read-only ergonomics to `Pcf`: `__repr__` → `__str__`, a `.dtype` property, `.times`/`.values`/`.breakpoints`/`.t_min`/`.t_max` accessors derived from `to_numpy()`, and a `Pcf.from_arrays(times, values, dtype=None)` classmethod. I tried to break it; the change is clean and the implementation matches the description. The accessors and `from_arrays` correctly delegate to the existing constructor so all breakpoint validation still runs. I found no correctness bugs — one edge case worth confirming and a couple of nits.

### 🟡 Edge cases to confirm

- **`from_arrays([], [])` silently fabricates a `(0, 0)` breakpoint instead of raising.** In `stablebear/functional/pcf.py`, `from_arrays` passes the 1-D guard (both ndim 1) and the length guard (both length 0), then `np.column_stack(([], []))` yields a `(0, 2)` array. That array reaches the C++ constructor, where `src/python/pypcf_support.hpp:42-46` special-cases `buf.size == 0` by emplacing a single `(0, 0)` point and returning — *before* the ndim/shape checks. Net effect: `Pcf.from_arrays([], [])` does not error and returns a one-breakpoint PCF the user never specified. This is pre-existing constructor behavior, but `from_arrays`'s docstring advertises validation ("Raises ValueError ... if the breakpoints are invalid"), so an empty input arguably should be rejected (or at least documented). There is no test covering empty input; consider adding one to pin down whatever the intended contract is.

- **`t_min` docstring says "always `0`".** This holds because the constructor pins `t_0 = 0` (`pypcf_support.hpp:84`), including the empty→`(0,0)` path above. Accurate, but worth a glance to make sure the doc and the empty-input behavior stay consistent if the latter is changed.

### 🟢 Nits

- **`.times` / `.values` return writable, non-contiguous views into a fresh temporary `to_numpy()` buffer.** Each property call materializes a new array via `np.asarray(self._data)` and slices `[:, 0]` / `[:, 1]`, so there's no cross-call aliasing and no shared-state risk. But the returned slice is technically writable and not contiguous; callers who do `f.times[0] = ...` would mutate a throwaway buffer with no effect on the PCF — slightly surprising for "read-only accessors." Not a bug; optionally return a copy or set `WRITEABLE=False` for clarity.

- **Test coverage is otherwise good and genuinely exercises the fixes.** `TestFromArrays` asserts the real error messages (`"same length"`, `"1-D"`, and the C++ `"t=0"` message at `pypcf_support.hpp:87`), `test_non_1d_values_raises` correctly checks that the guard covers the `values` argument too, and integer-dtype round-tripping is verified for both `pcf32i`/`pcf64i`. No weak/wrong-assertion tests spotted; the only gap is the empty-input case above.

**Verdict:** Solid, low-risk additive PR. No likely bugs; the only thing I'd nail down before merge is the empty-input contract for `from_arrays`.

---

PR #145, opened 2026-06-26
Validate arguments to system runtime setters (#101)

### LINKED ISSUE #101: [friction/interop] Negative / zero arguments to system setters raise leaky pybind/taskflow errors instead of clean

### Scenario
A user calls set_cuda_threshold(-1) (perhaps to mean 'never use GPU') or limit_cpus(0).

### Current approach
```
set_cuda_threshold(-1) -> 'TypeError: set_cuda_threshold(): incompatible function arguments...' (C++ takes an unsigned type, so a negative Python int is rejected with the overload-table dump). limit_cpus(0) -> 'RuntimeError: [/workspaces/masspcf/3rd/taskflow/.../executor.hpp:1215] executor must define at least one worker' (a taskflow assertion leaking an internal source path). The docstrings document the params only as 'n: int' with no stated lower bound.
```

### Desired API
Validate in the Python wrappers (system.py) and raise a clear ValueError, e.g. 'set_cuda_threshold: n must be >= 0' and 'limit_cpus: n must be >= 1', and document the valid range in the docstrings -- rather than leaking pybind signature dumps or taskflow assertions.

### Notes
All other system getters/setters round-trip fine (get_parallel_eval_threshold after set works; build_type()=='CUDA'; huge values up to int64 accepted; >int64 cleanly rejected as overflow).

bwehlin:
## Summary

The runtime-config setters in `stablebear/system.py` forwarded their integer argument straight to the C++ backend with **no validation**. This leaked implementation details:

- A **negative** argument triggered pybind's `TypeError: ... incompatible function arguments` overload dump (the C++ params are unsigned `size_t` / `unsigned int`).
- A **zero** worker count made taskflow raise a `RuntimeError` referencing an internal source path (`.../3rd/taskflow/.../executor.hpp:1215] executor must define at least one worker`).

The Python wrappers now validate up front and raise a clean `ValueError` naming the parameter and its bound (mirroring the existing `norms.py` pattern `raise ValueError("p must be >= 1.")`):

| Setter | Bound |
| --- | --- |
| `limit_cpus(n)` | `n >= 1` (0-worker pool invalid; use `force_cpu()` for CPU-only) |
| `limit_gpus(n)` | `n >= 1` |
| `set_cuda_threshold(n)` | `n >= 0` |
| `set_parallel_eval_threshold(n)` | `n >= 0` |
| `set_min_block_side(n)` | `n >= 0` |
| `set_block_size(x, y)` | `x >= 1`, `y >= 1` |

Non-numeric setters (`force_cpu`, `set_device_verbose`) are unchanged.

## Tests

New `test/python/test_system_validation.py` exercises the **error paths only** (validation runs before the C++ call, so rejected calls do not mutate global runtime state). It asserts each message does not contain `incompatible function arguments` / an internal source path and does name the parameter and bound. Full Python suite passes (1810 passed, 0 failed).

Docstrings and `docs/gpu.rst` note the bounds; `CHANGELOG.md` updated.

Closes #101


bwehlin:
## 🔍 Adversarial review

**Summary.** This PR adds up-front bounds validation to the runtime-config setters in `stablebear/system.py`, replacing leaky pybind `TypeError` overload dumps / internal taskflow `RuntimeError`s with clean `ValueError`s naming the parameter and bound. I cross-checked the wrappers against the C++ bindings (`src/python/pymodule.cpp`) and executor (`include/sbear/executor.hpp`). The change is small, correct for its stated scope, and the existing setter call sites (`conftest.py`, `test_pcf_tensor_eval.py`, `test_cuda_pdist_kernel.py`) all use values still permitted by the new bounds, so no regression. The bounds themselves are well-justified: `limit_cuda_workers(0)` and `set_block_dim(0, …)` would *not* be caught in C++ (they'd silently build a zero-worker pool / zero-dim block), so rejecting them in Python is a genuine improvement, not just cosmetic.

### 🟡 Edge cases to confirm

* **Non-integral / non-numeric arguments still leak the pybind dump.** The guards only check `n < 1` / `n < 0`, which doesn't constrain *type*. A float that passes the bound — e.g. `limit_cpus(2.5)`, `set_cuda_threshold(1.5)`, `set_block_size(16.0, 16)` — sails through validation and then hits `cpp.limit_cpus(2.5)` etc., where the `size_t` / `unsigned int` param raises exactly the `TypeError: ... incompatible function arguments` overload dump this PR set out to eliminate. Same for a string/`None`. If the intent of #101 is "no leaky backend errors from these setters," that goal is only met for the negative/zero case; non-integral inputs are still leaky. Worth either documenting the scope explicitly or adding an `isinstance(n, int)` (note: `bool` is an `int` subclass, so decide whether `limit_cpus(True)` should be allowed) check. Low severity since negatives/zero were the reported symptom, but it's a real residual gap in the stated guarantee.

### 🟢 Nits

* **Tests never exercise the leaky path, so they can't regression-guard the message.** `test_system_validation.py` deliberately only hits the Python-side `ValueError` (good rationale: avoids mutating global runtime state). But that means the `_LEAKY` substring assertions (`"incompatible function arguments"`, `"taskflow"`, `".hpp"`, …) are checking strings that the new `ValueError` messages could never contain anyway — they don't actually prove the *backend* error was suppressed, only that the hand-written message is clean. The tests correctly assert the new behavior; just note they wouldn't catch a regression where validation is bypassed and the float/type path re-leaks (see edge case above). Consider one test that asserts a non-integral arg is also rejected cleanly, which would close both gaps at once.
* **`bool` argument message is slightly odd.** `limit_cpus(False)` → `ValueError("limit_cpus: n must be >= 1 (got False)")`. Harmless, but `False`/`True` rendering in the bound message is a minor wart if type-checking is added later.
* **Adjacent (not introduced by this PR):** `cpp.limit_gpus` is bound only under `#ifdef BUILD_WITH_CUDA` (`pymodule.cpp:147`), so on a CPU-only build a *valid* `system.limit_gpus(n)` raises `AttributeError` rather than a clean message. This PR's `n >= 1` guard runs first, so the new error-path tests are unaffected, but the valid path on CPU builds remains rough. Out of scope here; flagging only for awareness.

**Verdict:** Solid, well-targeted fix with appropriate bounds and no regressions to existing call sites. The one substantive thing I'd address before merge is the non-integral-argument gap, since it leaves the exact leaky `TypeError` the PR aims to remove still reachable.

---

PR #146, opened 2026-06-26
squeeze/expand_dims accept a tuple of axes; squeeze no-ops on 0-d (#70, #71)

### LINKED ISSUE #70: [friction/shapeops] squeeze and expand_dims do not accept a tuple of axes (NumPy does)

### Scenario
User wants to drop two specific size-1 axes at once, or insert several new axes, the standard NumPy way: a.squeeze((0, 2)) or np.expand_dims(a, (0, 1)).

### Current approach
```
ts.squeeze((0,2)) raises TypeError (pybind: axis must be SupportsInt|SupportsIndex). te.expand_dims((0,1)) raises TypeError. Workaround: chain single-axis calls in the right order while accounting for shifting indices, e.g. ts.squeeze(2).squeeze(0) and te.expand_dims(0).expand_dims(0) -- error-prone because each call renumbers the remaining axes.
```

### Desired API
squeeze(axis=None | int | tuple[int,...]) and expand_dims(axis: int | tuple[int,...]) matching numpy.squeeze / numpy.expand_dims, with negative-axis support.

### Notes
NumPy: np.arange(6).reshape(1,6,1).squeeze((0,2)).shape == (6,); np.expand_dims(np.arange(6),(0,1)).shape == (1,1,6).

### LINKED ISSUE #71: [friction/shapeops] squeeze(0) on a 0-d tensor raises instead of returning the 0-d tensor (NumPy parity gap)

### Scenario
Generic code calls x.squeeze(0) on tensors of arbitrary rank, including the 0-d case produced by full-integer reductions/reshape(()).

### Current approach
```
t0 = FloatTensor(np.array(3.0)); t0.squeeze(0) raises 'ValueError: squeeze: axis 0 out of range for tensor with 0 dimensions'.
```

### Desired API
Match NumPy: np.array(3.0).squeeze(0) returns a 0-d array (shape ()). squeeze(0) on a 0-d tensor should be a no-op returning the 0-d tensor.

### Notes
Minor; only affects generic axis-parametric code that hits the 0-d edge.

bwehlin:
## Summary

`Tensor.squeeze` and `Tensor.expand_dims` previously only accepted a single integer axis (delegating straight to `operator.index`), so passing a tuple/list raised a raw `TypeError`. And `squeeze(0)` on a 0-d tensor raised `IndexError` instead of being a no-op like NumPy.

Both methods now accept a tuple/list of axes (NumPy parity), implemented in Python by chaining the existing single-axis C++ views:

- **`squeeze(axis)`** with a tuple/list: resolves each axis, rejects duplicates and non-size-1 axes with `ValueError` *before* mutating, then removes axes in descending order so earlier removals don't renumber the rest.
- **`expand_dims(axis)`** with a tuple/list: resolves each axis against the output rank (`ndim + len(axis)`), rejects duplicates, and inserts them in ascending order to match `numpy.expand_dims`.
- **0-d `squeeze`**: `squeeze(0)` / `squeeze(-1)` on a 0-d tensor now returns it unchanged; any other axis still raises `IndexError`.

The single-int paths are unchanged.

## Tests

Added tuple/list, negative-axis, duplicate-axis, non-size-1, and 0-d cases to `TestSqueeze`/`TestExpandDims` in `test/python/test_tensor_reshape.py`, comparing against NumPy references. Full Python suite: 1843 passed.

Docs (`docs/tensors.rst`) and `CHANGELOG.md` updated.

Closes #70
Closes #71


bwehlin:
## 🔍 Adversarial review

### Summary
This PR extends `Tensor.squeeze` / `Tensor.expand_dims` to accept a tuple/list of axes by chaining the existing single-axis C++ views in Python, and makes `squeeze(0)`/`squeeze(-1)` a no-op on 0-d tensors. I tried hard to break it against the documented NumPy semantics; the logic is correct on every edge case I could construct, and the tests genuinely exercise the new behavior (they assert both values and shapes against NumPy references). **Verdict: solid, ready to merge.** Only minor nits below.

### Findings

I specifically probed and confirmed the following all behave correctly (no bug):

- **0-d tuple vs scalar distinction.** `squeeze(0)` on a 0-d tensor is special-cased to a no-op, but a tuple like `squeeze((0,))` on a 0-d tensor is *not* — it falls into the tuple path, `_resolve_axis(0, 0)` raises `IndexError`. This exactly matches NumPy, where `np.array(3).squeeze(0)` returns `()` but `np.array(3).squeeze((0,))` raises an `AxisError` (which is an `IndexError` subclass). Nice subtlety to get right.
- **Negative-axis duplicate collisions.** `squeeze((0, -3))` on shape `(1,6,1)` resolves both to axis 0 and raises `ValueError` — matches NumPy's "duplicate value in 'axis'". For `expand_dims`, axes are correctly resolved against `out_ndim = ndim + len(axis)` *before* the duplicate check, mirroring `numpy.core.normalize_axis_tuple`, so e.g. `(0, -2)` on a 1-d input resolves to `(0, 1)` (no false-positive duplicate).
- **Removal/insertion order.** Descending removal for squeeze and ascending insertion for expand_dims (after resolving to *final* positions) match NumPy; the `test_squeeze_tuple_unsorted_order` / `descending_removal_order` cases lock this in.
- **Empty tuple.** `squeeze(())` and `expand_dims(())` both fall through as no-ops, matching NumPy (not tested, but correct — see nit).
- **Error message parity.** The non-size-1 message is copied verbatim from NumPy ("cannot select an axis to squeeze out which has size not equal to one").
- **Exception type for 0-d out-of-range.** `squeeze(1)` on 0-d → `IndexError`, consistent with `np.AxisError ⊂ IndexError`.

### 🟢 Nits
- **Non-tuple/list iterables for `axis`.** The `isinstance(axis, (tuple, list))` gate means a NumPy array or other iterable of axes (which real NumPy accepts) falls through to `operator.index(...)` and raises `TypeError`. This is consistent with the previous single-int-only behavior and outside the PR's stated scope, so not a regression — just noting it if full NumPy parity is ever desired.
- **No empty-tuple test.** `squeeze(())` / `expand_dims(())` are correct no-ops but untested. A one-line case would pin the behavior, since it's an easy thing to break in a future refactor of the loop.
- Both nits are optional; nothing blocks merge.


---

PR #147, opened 2026-06-26
load reports the real error on a corrupt tensor file (#38)

### LINKED ISSUE #38: [io] mpcf.load surfaces a misleading "Expected format type SingleObject ... but got SingleTensor" error

### Summary
io._load_any tries _load (tensor path) first; on its RuntimeError it falls through `except RuntimeError:` to _load_object. For a truncated tensor, _load_object re-reads the header, sees FormatType::SingleTensor, and raises the SingleObject mismatch error, which then becomes the user-visible exception -- masking the original 'corrupted/EOF' tensor error. The fallback is meant to distinguish object vs tensor files, but it swallows genuine tensor-parse failures and substitutes a wrong diagnosis.

### Minimal reproduction
```python
import io
import numpy as np
import masspcf as mpcf

X = mpcf.random.noisy_sin((3,), dtype=mpcf.pcf64)
buf = io.BytesIO(); mpcf.save(X, buf); good = buf.getvalue()
trunc = good[:200]                      # truncated tensor file (full is 1081 bytes)
try:
    mpcf.load(io.BytesIO(trunc))
except Exception as e:
    print('load() says:', type(e).__name__, '->', e)
    # RuntimeError -> Expected format type SingleObject for this operation but got format type SingleTensor

from masspcf.io import _load
try:
    _load(io.BytesIO(trunc))
except Exception as e:
    print('real cause:', type(e).__name__, '->', e)
    # RuntimeError -> Incorrect number of bytes returned (file may be corrupted)
```

### Observed
Running from /tmp: mpcf.load(io.BytesIO(trunc)) raises RuntimeError: "Expected format type SingleObject for this operation but got format type SingleTensor". This is the opposite of the truth -- the truncated bytes ARE a SingleTensor file. Calling masspcf.io._load directly on the same bytes raises the genuine cause: RuntimeError: "Incorrect number of bytes returned (file may be corrupted)". Exit code 0 (clean raise, no crash). The misleading message wins because of the fallback control flow in _load_any.

### Expected
A truncated/corrupt tensor file should surface an error reflecting the real cause (e.g. "file may be corrupted" / EOF), not one claiming the format type is SingleObject. The load() docstring promises to return the saved item or raise; the raise is honored (no crash), but the diagnostic actively misdirects -- it tells the user the file's format type is SingleObject when the file genuinely begins with a valid SingleTensor format-type marker and only fails later during tensor-body parsing.

### Root-cause hint
masspcf/io.py:_load_any -- the broad `except RuntimeError: ... return _load_object(buf)` fallback cannot distinguish "not a tensor file" (a FormatType mismatch) from "is a tensor file but corrupt/truncated". A truncated SingleTensor's real error from _load (read_bytes in include/mpcf/io/io_stream_base.hpp:85) is swallowed; _load_object then re-reads the header, sees FormatType::SingleTensor != SingleObject, and throws the mismatch error (include/mpcf/io.hpp:301), which becomes the user-visible exception. Fix: peek the FormatType field once and dispatch on its actual value instead of try/except fallback, so the correct path's corruption error propagates.

In `test/python/test_bugscan_io.py` -- run with `cd test && python -m pytest python/test_bugscan_io.py`

```python
def test_load_truncated_tensor_error_reflects_corruption():
    """A truncated tensor file should surface the real corruption cause.

    Does not crash (clean RuntimeError, exit 0), so this is a normal in-process
    test.
    """
    # BUG: #37 load() reports the wrong format type for a corrupt tensor file
    # Expected: the surfaced error reflects the real cause (the file is a
    #           corrupt/truncated SingleTensor -> "file may be corrupted"/EOF),
    #           NOT a claim that the format type is SingleObject.
    # Observed today: RuntimeError "Expected format type SingleObject for this
    #                 operation but got format type SingleTensor" -- the exact
    #                 opposite of the truth, masking the genuine parse error.
    X = mpcf.random.noisy_sin((3,), dtype=mpcf.pcf64)
    buf = io.BytesIO()
    mpcf.save(X, buf)
    good = buf.getvalue()
    trunc = good[:200]  # truncated SingleTensor file (full is ~1081 bytes)

    with pytest.raises(Exception) as excinfo:
        mpcf.load(io.BytesIO(trunc))

    msg = str(excinfo.value)
    # The diagnostic must not misdirect the user about the file's format type.
    assert "Expected format type SingleObject" not in msg, (
        "load() surfaced the misleading SingleObject mismatch error for a "
        f"corrupt SingleTensor file; message was: {msg!r}"
    )
    # And it should point at the real cause (truncation/corruption/EOF).
    assert any(
        token in msg.lower() for token in ("corrupt", "byte", "eof", "truncat")
    ), f"load() error should reflect corruption; message was: {msg!r}"
```

</details>

bwehlin:
## Summary

Closes #38.

`sb.load` of a truncated/corrupt single-tensor file surfaced a misleading error:

> Expected format type SingleObject ... but got format type SingleTensor

Root cause: `_load_any` (in `stablebear/io.py`) tried the tensor path and, on **any** `RuntimeError`, blindly retried the single-object path. For a truncated tensor the real tensor-load error ("Incorrect number of bytes returned (file may be corrupted)") was swallowed, and the object path then raised the misleading format-type message because the header still said `SingleTensor`.

## Fix (Python-only, no rebuild)

In the `except RuntimeError` branch, only fall back to `_load_object` when the tensor-load error message contains the marker `format type SingleObject` — i.e. the C++ reader confirmed the file genuinely is a `SingleObject` ("...but got format type SingleObject"). For any other `RuntimeError` (genuine corruption/EOF), re-raise the original error so the real cause surfaces.

## Tests

New `test/python/test_bugscan_io.py`:
- Truncated tensor `load` raises an error that does NOT contain "format type SingleObject" and DOES mention corruption/bytes/EOF/truncation.
- Valid single objects (`Pcf`, `DistanceMatrix`, `Barcode`) still round-trip via the object fallback.
- A valid tensor still round-trips.
- Empty bytes / random garbage still raise.

Full Python suite: 1807 passed.


bwehlin:
## 🔍 Adversarial review

### Summary
The PR narrows `_load_any`'s blind `_load_object` fallback in `stablebear/io.py` so it only retries the single-object path when the tensor-load `RuntimeError` contains the marker `"format type SingleObject"`, otherwise re-raising the real error. I traced the change against the C++ reader and the claim holds: the fix is correct, minimal, and backed by tests that genuinely exercise it. **Verdict: solid, recommend merge** (with one fragility nit below).

I verified the control flow end-to-end:
- A truncated *tensor* file keeps its header + `SingleTensor` format byte, so `read_any_tensor` (`include/sbear/io.hpp:226-231`) passes the format-type check and then fails deep in the payload at `read_bytes` with `"Incorrect number of bytes returned (file may be corrupted)"` (`include/sbear/io/io_stream_base.hpp:71`). That message has **no** `"format type SingleObject"` substring, so the new branch re-raises it — exactly the bug fix.
- A *genuine object* file makes `read_any_tensor` raise `"Expected format type SingleTensor ... but got format type SingleObject"`. That string **does** contain `"format type SingleObject"`, so the fallback to `_load_object` still fires. Verified the substring match is satisfied by the "but got" clause — the fallback for real objects is preserved.
- pybind11 maps `std::runtime_error` → Python `RuntimeError` with the message intact (`src/python/py_io.cpp:25-30`), so the keyword/substring assertions are well-founded.

### 🟢 Nits
- **Brittle coupling to a C++ message string** (`stablebear/io.py`, new `except` branch). The dispatch now depends on the exact substring `"format type SingleObject"` produced in `include/sbear/io.hpp:289`. If that exception text is ever reworded, the Python fallback silently breaks and *valid* single objects would stop loading — a silent, hard-to-diagnose regression. Consider asserting this contract in a test (a direct check that a real object's tensor-path error contains the marker) or, longer-term, surfacing the format type as structured data rather than parsing the message. Acceptable given the stated Python-only / no-rebuild constraint.
- **`test_empty_bytes_still_raises` / `test_random_garbage_still_raises`** don't actually exercise the new branch — both fail at `read_header` ("Incorrect number of bytes" / "Unrecognized file format.", no marker) and re-raise, never reaching `_load_object`. They're fine as regression guards but, despite their placement in this file, they don't cover the fallback logic; the truncated-object case (object file truncated *after* its `SingleObject` byte, which would route through the fallback and then fail) is the untested sibling of the new path. Not required, but it's the one branch the suite leaves uncovered.

### Notes on robustness (no action needed)
- The last-8-bytes truncation in `test_truncated_tensor_surfaces_real_error_not_misleading_one` reliably corrupts the payload tail (one trailing `float64` of PCF data) after the format bytes, so it deterministically hits the corruption path rather than the format-type check — a correct, non-flaky construction.
- Corrupt-header cases in object files cause no regression: both old and new code raise the same `read_header` error.

No correctness bugs found. The fix does what the description claims and the core regression test asserts the right thing.

---

PR #148, opened 2026-06-26
SymmetricMatrix/DistanceMatrix: shape, len, diagonal, row/slice access (#79)

### LINKED ISSUE #79: [friction/kernel] SymmetricMatrix has no shape/len/row access/slicing - extracting a row or diagonal needs a

### Scenario
After l2_kernel, a user wants the diagonal (self-similarities), a single row, or just the matrix dimension without converting to a full dense array.

### Current approach
```
K = mpcf.l2_kernel(X)
len(K)          # TypeError: object of type 'SymmetricMatrix' has no len()
K.shape         # AttributeError: no attribute 'shape'
K.diagonal()    # AttributeError: no attribute 'diagonal'
K[0]            # TypeError: cannot unpack non-iterable int object
K[0:2, 0]       # TypeError: '<' not supported between 'slice' and 'int'
# Only working paths: K.size (n), and an explicit double-index loop:
row0 = [K[0, j] for j in range(K.size)]
diag = [K[i, i] for i in range(K.size)]
```

### Desired API
Add SymmetricMatrix.shape -> (n, n), __len__ -> n, .diagonal() -> np.ndarray, and support row/slice indexing (K[i] returns row i as a 1-D array, K[i, :] / K[:, j] / K[a:b, c:d] return sub-arrays) so callers don't drop to per-element loops or to_dense() for simple slices.

### Notes
K.size and K.storage_count exist; the gap is shape/len and any vectorized row/diagonal/slice access.

bwehlin:
Closes #79.

`SymmetricMatrix` and `DistanceMatrix` (the scalar compressed-matrix classes) previously exposed only `(i, j)` scalar element access — no `shape`, `len`, `diagonal`, or row/slice indexing.

## Changes (Python-only, no rebuild)
Added to both classes:
- `shape` property -> `(n, n)`.
- `__len__` -> `n`.
- `diagonal()` -> 1-D NumPy array. `SymmetricMatrix` returns the stored diagonal; `DistanceMatrix` returns zeros of the matching float dtype (its diagonal is always zero).
- `__getitem__` now delegates any non-`(int, int)` index (bare `int`, `slice`, or a tuple containing a slice / of length != 2) to `self.to_dense()[index]`, giving NumPy row/column/sub-array semantics and negative indexing. The fast `(i, j)` int-pair path is unchanged and still returns the scalar entry. `__setitem__` is unchanged (int-pair only).

Empty (n=0) matrices give `shape == (0, 0)`, `len == 0`, and an empty `diagonal()`.

## Tests, docs, changelog
- Added `TestShapeLenDiagonal` and `TestRowSliceIndexing` to `test_symmetric_matrix.py` and `test_distance_matrix.py` (both dtypes; includes empty-matrix edge cases and scalar-pair preservation).
- Documented the new accessors in `docs/distances.rst`.
- Added a changelog bullet under `0.4.4` -> Bug fixes.

Full Python suite: 1839 passed, 0 failures.


bwehlin:
## 🔍 Adversarial review

### Summary
The PR adds `shape`, `__len__`, `diagonal()`, and NumPy-style row/slice indexing to the scalar `SymmetricMatrix` and `DistanceMatrix` classes, keeping the fast `(int, int)` scalar path and delegating everything else to `to_dense()[index]`. The change is Python-only, well-scoped, and the new tests genuinely exercise the new behavior (both dtypes, empty matrices, negative indices, scalar preservation, `np.integer` pairs, out-of-range). Overall this is a solid PR; I found one real dtype inconsistency and a couple of nits, no correctness bugs in the indexing logic itself.

### 🟡 Edge cases to confirm

**1. `SymmetricMatrix.diagonal()` does not preserve the matrix dtype (returns `float64` for a `float32` matrix).**

In `stablebear/symmetric_matrix.py`:
```python
def diagonal(self) -> np.ndarray:
    return np.array([self._data[i, i] for i in range(self.size)])
```
The C++ binding's `__getitem__` (`src/python/py_symmetric_matrix.hpp`, the `self(ij.first, ij.second)` lambda) returns a `T` that pybind11 converts to a **Python `float`**. So `np.array([...])` of Python floats is always `float64`, regardless of whether the matrix is `float32`. This is inconsistent with:
- `to_dense()`, which returns `float32` for a float32 matrix, and
- `DistanceMatrix.diagonal()` in this same PR, which deliberately matches dtype via `np.float32 if self.dtype == float32 else np.float64`.

The new `test_diagonal` only uses `np.testing.assert_array_equal` (values), so it does not catch the dtype mismatch. Suggest building the array with the matching float dtype (e.g. `np.fromiter(..., dtype=np_float, count=self.size)` or `np.array([...], dtype=np_float)`) and asserting `diag.dtype` in the test, mirroring the `DistanceMatrix` side.

**2. Empty `SymmetricMatrix.diagonal()` dtype, same root cause.**

For `n == 0`, `np.array([])` is `float64` even for a float32 matrix. `test_empty_shape_len_diagonal` only asserts `.shape == (0,)`, so this slips through too. Fixing (1) with an explicit dtype resolves this as well.

### 🟢 Nits

- **`bool` is accepted into the scalar `(i, j)` fast path.** The guard `all(isinstance(k, (int, np.integer)) for k in index)` treats `m[True, False]` as `m[1, 0]` (Python `bool` is an `int` subclass). Harmless in practice and arguably matches no sensible alternative, but worth noting since NumPy gives booleans special (mask) semantics for the dense path.
- **Every slice/row/column access materializes the full dense `n×n` matrix.** `m[:, j]` and `m[i]` both call `to_dense()`, defeating the compressed-storage memory savings for large matrices. This is a reasonable simplicity tradeoff and the docstring states the dense semantics, so it's documentation-accurate — just flagging it as a potential follow-up for hot paths.

### Verdict
The indexing dispatch (int-pair fast path vs. dense delegation), `shape`, `__len__`, and `DistanceMatrix.diagonal()` are all correct and well-tested. The only substantive item is the `SymmetricMatrix.diagonal()` dtype inconsistency (finding 1/2), which is low-severity but a genuine behavior gap relative to the rest of the API and not covered by the tests.

---

PR #149, opened 2026-06-26
Broadcast a size-1 tensor RHS on masked/index assignment (#60)

### LINKED ISSUE #60: [friction/setitem] Masked/index/outer tensor-RHS assignment does not broadcast a scalar-shaped (size-1) tensor,

### Scenario
User wants to assign a single tensor value to all positions selected by a boolean mask or an integer index array, e.g. set the masked region to the value held in a 1-element tensor (a reduction result, a computed constant kept as a tensor). In NumPy `a[mask] = np.array([99.])` and `a[[1,3]] = np.array([99.])` broadcast the size-1 RHS across all selected positions.

### Current approach
```
from masspcf.tensor import FloatTensor, BoolTensor, IntTensor
a = FloatTensor(np.arange(6, dtype=np.float32).reshape(2,3))
m = BoolTensor(np.array([[True,False,True],[False,True,False]]))
a[m] = FloatTensor(np.array([99.], dtype=np.float32))
# -> ValueError: masked_assign: more true values in mask than elements in values
b = FloatTensor(np.arange(5, dtype=np.float32))
b[IntTensor(np.array([1,3]))] = FloatTensor(np.array([99.], dtype=np.float32))
# -> ValueError: index_assign: values shape does not match expected shape
# Workaround: drop to a Python float scalar (a[m] = 99.0) which routes to masked_fill and works,
# OR manually count the selected positions and tile the RHS to that exact length first.
```

### Desired API
masked_assign / index_assign / axis_assign / outer_assign should broadcast a size-1 (or otherwise broadcast-compatible) tensor RHS across the selected region, matching the __setitem__ docstring's claim that a tensor RHS is 'broadcast to the selected shape' and matching NumPy. I.e. `a[mask] = FloatTensor([99.])` should fill every masked position with 99.

### Notes
The scalar (non-tensor) path works because it routes to masked_fill/index_fill; only the Tensor-RHS path fails to broadcast. The __setitem__ docstring explicitly says a tensor RHS is 'broadcast to the selected shape', so this is a doc/behavior gap as much as friction.

bwehlin:
## Summary

Fixes #60: a size-1 tensor right-hand side did not broadcast on masked/index/outer assignment.

For a `Tensor` RHS, `__setitem__` routed through the C++ `*_assign` helpers (`masked_assign` / `index_assign` / `axis_assign` / `outer_assign`), which require an exact element-count/shape match. So `a[mask] = sb.FloatTensor([99.])` raised e.g. `masked_assign: more true values in mask than elements in values` instead of broadcasting the single value, unlike NumPy (where `arr[mask] = np.array([99.])` fills every selected position).

## Fix

In `_setitem_entries` (`stablebear/_tensor_base.py`), if the RHS is a `Tensor` with `size == 1`, extract its lone element (after dtype coercion via `_coerce_rhs`) so it flows through the broadcasting `*_fill` path. The multi-element tensor path is unchanged, so genuine length/shape mismatches still raise. Dtype validation still runs up front in `__setitem__`, so a wrongly-typed RHS is still rejected.

This covers masked, integer-index, and outer (multi-axis) assignment, plus basic-slice assignment.

## Tests

Added `TestSize1TensorRhsBroadcast` in `test/python/test_tensor_mask.py`:
- boolean-mask assign of a size-1 `FloatTensor` fills all selected positions (vs NumPy reference)
- integer-index assign of a size-1 `FloatTensor`
- outer (multi-axis) assign of a size-1 `FloatTensor`
- 0-d RHS broadcasts uniformly (value-agnostic; 0-d numpy readback is unreliable)
- all-False mask with a size-1 RHS writes nothing
- multi-element size mismatch still raises

Full Python suite: 1805 passed, 0 failures.

Closes #60


bwehlin:
## 🔍 Adversarial review

### Summary

This PR fixes #60 by broadcasting a size-1 `Tensor` RHS like a scalar on masked/index/outer/basic-slice assignment. The fix is placed at a single chokepoint at the top of `_setitem_entries` (`stablebear/_tensor_base.py`), which is a clean design — every advanced-indexing path inherits the broadcast uniformly, and genuine multi-element shape mismatches still fall through to the count-checking `*_assign` path. I tried hard to break the core numeric path and could not: **the claimed functionality looks correct**. The findings below are edge cases and test-quality nits, not blockers. (Note: I could not build the extension in this environment — submodule fetch is blocked — so this is a static analysis cross-checked against the local tree.)

### Findings

#### 🟡 Edge cases to confirm

- **size-1 `FloatTensor` RHS into a `PointCloudTensor`** (`_tensor_base.py`, new branch ~L518-525). For a `PointCloudTensor`, `_coerce_rhs` returns the `FloatTensor` unchanged (it isn't a `NumericTensor`), so `coerced._represent_element(...)` uses **`FloatTensor`'s identity** `_represent_element` and decays the RHS to a bare Python `float`. That then routes into the scalar cloud-fill path in `_basic_setitem` / `masked_fill`, where a single float is interpreted as a "cloud". This has no clear NumPy analog and is untested; worth confirming it either errors cleanly or is intentionally out of scope (it likely never worked before either, so probably not a regression).

- **Matrix/PCF tensor RHS now broadcasts too.** Because the branch fires for *any* `Tensor` with `size == 1`, a size-1 `DistanceMatrixTensor` / `SymmetricMatrixTensor` / `PcfTensor` RHS (all valid setitem dtypes for their tensors) now also broadcasts via `_represent_element` → `DistanceMatrix(...)` / `Pcf(...)` → fill, where previously the `*_assign` path applied. This is a reasonable generalization, but it is a behavior change **beyond what the description and tests cover** (tests are Float-only). Worth a sentence in the changelog and at least one non-numeric test.

#### 🟢 Nits

- **`test_zero_d_rhs_broadcasts_uniformly` asserts too little.** The destination `t` is a 1-d size-5 `FloatTensor`, so `np.asarray(t)[mask]` readback is fully reliable — the "0-d numpy readback is unreliable" justification refers to the *RHS*, not the destination. The test only checks `np.all(filled == filled[0])`, so a mis-indexed-but-uniform extraction (e.g. decaying to `0.0`) would still pass. It can and should assert `filled == 99.0` like the other tests, which would actually pin down the value the 0-d branch extracts.

- **No outer (multi-axis) cross-dtype test.** `test_cross_dtype_size1_rhs_is_cast` covers the mask path and `test_cross_dtype_size1_rhs_int_index` the index path, but the outer/`outer_fill` path with a size-1 `IntTensor`→`FloatTensor` RHS is untested. Cheap to add for symmetry.

### Verdict

Solid, minimal, well-targeted fix with good tests for the numeric case. No correctness bugs found in the claimed behavior; the items above are coverage gaps and an over-broad-but-harmless generalization, not defects.


---

PR #150, opened 2026-06-26
Out-of-range reduction dim raises numpy.AxisError (#82)

### LINKED ISSUE #82: [friction/reductions] No way to validate/clamp dim before calling, and no documented error contract for bad dim

### Scenario
A user writing a generic pipeline that reduces along a user-supplied dim wants to guard against a bad dim value before invoking the (potentially crashing) reduction.

### Current approach
```
No public helper exists to check 'is dim valid for this tensor'. The user must manually check 0 <= dim < len(A.shape) by hand for every reduction call, because passing an out-of-range dim silently returns a wrong shape or aborts the interpreter rather than raising. The docstrings/Reductions docs do not state what happens for an out-of-range dim.
```

### Desired API
mean/max_time should themselves raise numpy.AxisError on out-of-range dim (matching NumPy), and the docstrings should document that contract. Optionally a Tensor.normalize_axis(dim) / shape helper for generic code.

### Notes
This is the user-facing face of the critical dim-validation bug; even once crashes are fixed, the documented error contract is missing.

bwehlin:
## Summary

Closes #82.

An out-of-range reduction `dim` for `mean`/`max_time` raised a plain `IndexError`. It now raises `numpy.AxisError`, which subclasses **both** `IndexError` and `ValueError`, so existing `pytest.raises(IndexError)` / handler code keeps working while matching NumPy's axis-error contract.

## Changes

- `stablebear/reductions.py`: `_resolve_dim` now raises `numpy.AxisError(dim, ndim)` instead of `IndexError`. Resolves the class via `np.exceptions.AxisError` (NumPy 2) with a fallback to `np.AxisError` (NumPy 1). Negative-dim normalization is unchanged. Updated the `_resolve_dim` docstring and the `Raises` sections of `mean`/`max_time`.
- `docs/evaluation.rst`: out-of-range `dim` now documented as raising `numpy.AxisError` (a subclass of `IndexError`).
- `CHANGELOG.md`: new bug-fix bullet under `0.4.4`.
- `test/python/test_bugscan_reductions.py`: tests that too-large and too-negative dims raise `numpy.AxisError` for both reductions, that it stays catchable as `IndexError` (backward compatibility), and a sanity check that valid negative dims still work. Existing `pytest.raises(IndexError)` reduction tests continue to pass.

## Verification

`cd test && python -m pytest python -q` -> 1808 passed, 0 failures.


bwehlin:
## 🔍 Adversarial review

### Summary
This PR changes `_resolve_dim` in `stablebear/reductions.py` so an out-of-range reduction `dim` raises `numpy.AxisError` instead of a plain `IndexError`, and adds tests asserting the new type while confirming backward-compatible `IndexError` catchability. The change is small, correct, and well-tested. I tried to break it and found no correctness bugs — only a couple of minor/cosmetic items below.

### Verification
I verified the load-bearing claims against numpy 2.4.6 and by simulating the patched `_resolve_dim`:
- `numpy.exceptions.AxisError` subclasses **both** `IndexError` and `ValueError` ✅ — so the existing `pytest.raises(IndexError)` tests (and `pytest.raises((IndexError, ValueError))`) keep passing.
- `AxisError(dim, ndim)` produces a sensible message for both positive and negative dims (`"axis 5 is out of bounds for array of dimension 2"`, `"axis -3 is out of bounds for array of dimension 2"`).
- Valid negative dims (`-1`, `-2`) still resolve correctly; the new `test_valid_negative_dim_unchanged_by_axis_error` genuinely guards against regression.
- `pytest.raises(_AxisError)` in the new test does require an actual `AxisError` (a plain `IndexError` would *not* satisfy it), so the test really exercises the claimed fix rather than passing vacuously.

### Findings

#### 🟢 Nits

1. **`reductions.py` (the `AxisError = getattr(np.exceptions, "AxisError", None) or np.AxisError` line) — the "NumPy 1" fallback is dead and self-contradictory.** The comment and PR description say the fallback supports NumPy 1, but `pyproject.toml` pins `numpy>=2.0` (twice). On NumPy 2 the top-level `np.AxisError` was **removed** (`hasattr(np, "AxisError")` is `False`), so:
   - `getattr(np.exceptions, "AxisError", None)` always succeeds on supported versions, making `or np.AxisError` unreachable.
   - If the first operand ever *were* `None` on NumPy 2, evaluating `np.AxisError` would itself raise `AttributeError` — i.e. the "fallback" can't actually fire safely. And on a hypothetical NumPy < 1.25, `np.exceptions` doesn't exist, so `getattr(np.exceptions, ...)` would raise `AttributeError` on attribute access before any fallback. Given the `>=2.0` floor, consider just `AxisError = np.exceptions.AxisError` with a comment, and drop the NumPy-1 framing from the comment/PR description.

2. **User-facing message wording changes from "dim" to "axis".** The old message was `"dim {dim} is out of range for tensor with {ndim} dimension(s)"`; `AxisError` emits `"axis N is out of bounds for array of dimension M"`. The public parameter is named `dim` and the rest of the API/docs speak of `dim`, so the new message ("axis", "array") is slightly off-vocabulary for stablebear. Harmless, but worth a note in the docstring/changelog or a custom message if message stability matters to anyone grepping logs.

3. **Redundant (harmless) assertion in the new test.** `assert issubclass(type(excinfo.value), IndexError)` inside `test_out_of_range_dim_raises_axis_error` can never fail given `pytest.raises(_AxisError)` already matched — every `AxisError` is an `IndexError`. It reads as documentation rather than a real check; fine to keep.

#### 🟡 Scope note (not a defect in this PR)
- Other axis-validation paths still raise plain `IndexError` — e.g. `stablebear/_tensor_base.py:144` (`"axis {axis} is out of range for tensor with {ndim} dimension(s)"`). After this PR, reductions and general tensor indexing are inconsistent about `AxisError` vs `IndexError`. That's legitimately outside issue #82's scope (which is about reductions), but if the goal is a uniform "axis-error contract" it's a natural follow-up.

### Verdict
Solid, correctly scoped fix with tests that actually exercise the new behavior and the backward-compat guarantee. No correctness, edge-case, or dtype bugs found. The only actionable item is the misleading/dead NumPy-1 fallback (Nit 1); everything else is cosmetic. 👍


---

Issue #83, opened 2026-06-07
[friction/persistence] Cannot build a BarcodeTensor from a list/array of numpy barcode arrays


bwehlin:
_Usability friction found by the masspcf API scan (2026-06-07). Not a correctness bug; see `friction-triage.md`._  **Priority: medium.**

### Scenario
A user has barcodes as numpy (n,2) arrays (e.g. loaded from disk, or computed elsewhere) and wants to assemble a BarcodeTensor to feed barcode_to_stable_rank / betti / apf in one parallel call.

### Current approach
```
mpers.BarcodeTensor([np.array([[0.,1.]]), np.array([[0.,2.]])]) raises AttributeError: 'numpy.ndarray' object has no attribute '_data'. The only working list-construction is from already-wrapped Barcode objects: mpers.BarcodeTensor([mpers.Barcode(a0), mpers.Barcode(a1)]). (Element assignment bcs[i] = ndarray does work, so the nested-construction path is inconsistent with element setitem, whose docstring lists np.ndarray as a valid dtype.)
```

### Desired API
BarcodeTensor([ndarray, ndarray, ...]) and BarcodeTensor(list_of_lists_of_ndarray) should accept raw (n,2) numpy arrays directly (wrapping them in Barcode internally), matching the documented setitem dtypes.

### Notes
ph_tensor.py _get_valid_setitem_dtypes lists np.ndarray, but _tensor_from_nested in the list-constructor only maps cpp Barcode types, so ndarrays fail.

Issue #80, opened 2026-06-07
[friction/kernel] Mixed-dtype list passed to l2_kernel raises an opaque internal pybind error instead of a clear


bwehlin:
_Usability friction found by the masspcf API scan (2026-06-07). Not a correctness bug; see `friction-triage.md`._  **Priority: low.**

### Scenario
A user assembles PCFs from different sources and accidentally mixes float32 and float64 Pcf objects, then calls mpcf.l2_kernel([p32, p64]).

### Current approach
```
p32 = mpcf.Pcf(np.array([[0,4],[3,0]], dtype=np.float32))
p64 = mpcf.Pcf(np.array([[0,2],[3,0]], dtype=np.float64))
mpcf.l2_kernel([p32, p64])
# TypeError: _set_element(): incompatible function arguments ... Invoked with:
#   <Pcf32Tensor object ...>, [1], <Pcf_f64_f64 object ...>
# Leaks the internal _set_element binding and tensor-construction details.
```

### Desired API
_to_tensor_pcf / PcfTensor construction (or l2_kernel) should detect heterogeneous element dtypes up front and raise a clear ValueError/TypeError like 'all PCFs must have the same dtype (got pcf32 and pcf64)', matching the documented dtype-mismatch error in _resolve_pcf_inputs.

### Notes
The mismatch is currently caught only deep in PcfTensor([...]) construction (it picks the first element's dtype, then fails to insert the rest), surfacing a low-level pybind error rather than the friendly message _resolve_pcf_inputs already documents for multi-input cases.

Issue #55, opened 2026-06-07
[friction/construction] IntTensor and PcfTensor raise leaky low-level errors (bare KeyError / pybind _set_element


bwehlin:
_Usability friction found by the masspcf API scan (2026-06-07). Not a correctness bug; see `friction-triage.md`._  **Priority: low.**

### Scenario
User passes an unsupported dtype to IntTensor, mixes pcf32 and pcf64 Pcf elements in one PcfTensor list, or passes plain floats/None where Pcf elements are expected.

### Current approach
```
mpcf.IntTensor([1,2], dtype=mpcf.float64)        # -> KeyError: masspcf.float64 (from convert dict lookup)
mpcf.PcfTensor([mk32, mk64])                     # -> TypeError: _set_element(): incompatible function arguments ... (raw pybind dump)
mpcf.PcfTensor([1.0, 2.0])                       # -> AttributeError: 'float' object has no attribute '_data'
mpcf.PcfTensor([mk, None])                       # -> AttributeError: 'NoneType' object has no attribute '_data'
```

### Desired API
Validate the dtype against the supported set up front and raise a descriptive TypeError (like typing._assert_valid_dtype does for zeros). For PcfTensor, detect mixed element dtypes and non-Pcf elements before calling into C++ and raise a clear message (the _tensor_from_nested 'Unsupported element type' path is unreachable because it dereferences flat[0]._data first).

### Notes
All of these correctly fail rather than crash, so they are friction (misleading/leaky errors), not bugs. The KeyError in particular surfaces an internal dict key as the exception message.

PR #151, opened 2026-06-26
Clean errors for tensor construction; BarcodeTensor from ndarray list (#54, #55, #80, #83)

### LINKED ISSUE #54: [friction/construction] FloatTensor silently ignores any dtype argument that is not exactly mpcf.float32

### Scenario
User asks for a specific element type, e.g. FloatTensor(data, dtype=np.float32) (numpy dtype) or accidentally FloatTensor(data, dtype=mpcf.int32).

### Current approach
```
mpcf.FloatTensor([1.0,2.0], dtype=np.float32).dtype   # -> float64 (not float32)
mpcf.FloatTensor([1.0,2.0], dtype=mpcf.int32).dtype    # -> float64, silently
mpcf.FloatTensor([1.0,2.0], dtype=mpcf.boolean).dtype  # -> float64, silently
The constructor only checks `if dtype == float32`; everything else falls through to float64 with no error.
```

### Desired API
FloatTensor should accept np.float32/np.float64 dtype aliases and should raise a clear TypeError for non-float dtypes (e.g. mpcf.int32, mpcf.boolean), rather than silently returning float64.

### Notes
Easy to get a silently-wrong precision. Contrast with IntTensor, which at least fails (see next item) instead of silently mis-typing.

bwehlin:
## Summary

Tensor construction used to surface leaky, low-level errors for bad inputs. This makes those paths raise clear, actionable `TypeError`s and lets `BarcodeTensor` be built from raw barcode arrays.

- **#54** — `FloatTensor(arr, dtype=...)` silently fell back to `float64` for any non-`float32` dtype (e.g. `sb.int32`). It now normalizes numpy aliases (`np.float32`, `np.dtype('float32')`) and rejects non-float dtypes with a clear `TypeError`. `FloatTensor(arr, dtype=np.float32).dtype == float32`.
- **#55a** — `IntTensor([...], dtype=sb.float64)` raised a bare `KeyError`. It now asserts a valid integer dtype and raises `TypeError`.
- **#55b / #80** — `_tensor_from_nested` leaked an `AttributeError` for a non-element item (e.g. a plain `float`) and dumped the raw pybind `_set_element` overload list for a mixed-precision list. Every flattened element is now validated up front: a non-element raises `TypeError` naming the offending type and flat index; a mixed-precision list raises a `TypeError` mentioning "same dtype" and the distinct element dtypes. This also fixes `l2_kernel([p32, p64])`, which routes through the same helper.
- **#83** — `BarcodeTensor([np.array([[0.,1.]]), ...])` raised `AttributeError`. `_tensor_from_nested` gained an optional `wrap=` callable; `BarcodeTensor` passes `wrap=Barcode`, so raw `(n, 2)` NumPy barcode arrays are wrapped automatically (idempotent for already-wrapped `Barcode` elements).

## Tests

- `test/python/test_int_tensor.py`: bad `dtype=` raises `TypeError`.
- `test/python/test_tensor_from_list.py`: `FloatTensor` dtype validation; PCF list with a plain-float element raises `TypeError` naming the type; mixed-precision `PcfTensor` and `l2_kernel` raise `TypeError` matching "same dtype"; `BarcodeTensor` from 1-D and 2-D lists of raw `(n, 2)` f32/f64 ndarrays yields the right shape and dtype.

Full Python suite: 1807 passed.

Docs: noted in `docs/tensors.rst` that `BarcodeTensor` accepts raw barcode arrays. Changelog updated under 0.4.4.

Closes #54, #55, #80, #83.


bwehlin:
## 🔍 Adversarial review

### Summary
The PR turns several leaky low-level errors (`KeyError`, `AttributeError`, raw pybind overload dumps) on bad tensor-construction inputs into clear `TypeError`s, and lets `BarcodeTensor` accept raw `(n, 2)` barcode ndarrays via a new `wrap=` hook in `_tensor_from_nested`. I traced every claimed path against the code and the changes do what they say — the `FloatTensor` dtype normalization, the `IntTensor` guard, the up-front per-element validation + "same dtype" mixed-precision check, the `l2_kernel` routing, and the `wrap=Barcode` idempotency all hold up. Overall this is a solid, well-tested change; the findings below are one edge case and minor consistency nits, not blockers.

### Findings

#### 🟡 Edge cases to confirm

**1. `BarcodeTensor` with a bad barcode leaf still surfaces an `AttributeError`, not a clean `TypeError` — and it's untested.**

In `_tensor_from_nested` (`stablebear/_tensor_base.py`), `wrap=Barcode` runs *before* the new `hasattr(elem, "_data")` validation:

```python
if wrap is not None:
    flat = [wrap(e) for e in flat]
# ... then the "element i has type X" / "same dtype" checks
```

But `Barcode.__init__` (`stablebear/persistence/barcode.py`) has a pre-existing bug: on its two `fail = True` paths (non-`float32/float64` ndarray, or a non-array/non-`Barcode` object) it **never assigns `self._data`**, then immediately does `if isinstance(self._data, cpp_p.Barcode32): ...` — which raises `AttributeError: 'Barcode' object has no attribute '_data'` *before* the intended `TypeError("Barcode cannot be constructed ...")` can fire.

Net effect: `sb.persistence.BarcodeTensor([np.array([1, 2, 3])])` (1-D / non-`(n,2)`), `BarcodeTensor([np.array([[0,1]], dtype=np.int64)])` (int dtype), or `BarcodeTensor([3.0])` (plain float leaf) still leak an `AttributeError` — exactly the class of error this PR set out to eliminate. The clean `"element {i} has type {...}"` message added to `_tensor_from_nested` is unreachable for `BarcodeTensor`, because `wrap` crashes first. The new tests only cover valid `float32`/`float64` `(n,2)` arrays (`test_barcode_tensor_from_ndarray_list_{1d,2d}`), so this hole isn't exercised. A targeted fix would be to assign `self._data = None` at the top of `Barcode.__init__` (or guard the `isinstance` checks), and add a `pytest.raises(TypeError)` test for a bad barcode leaf.

#### 🟢 Nits

**2. `FloatTensor` / `IntTensor` dtype-normalization asymmetry.** The new `FloatTensor` branch normalizes numpy aliases (`np.float32`, `np.dtype('float32')`, even strings) before validating, but the `IntTensor` branch (`base_tensor.py`) only adds `_assert_valid_dtype(dtype, [int32, int64, uint32, uint64])` with no normalization. So `FloatTensor(arr, dtype=np.float32)` works, while `IntTensor(arr, dtype=np.int32)` now raises `TypeError`. That's still an improvement over the prior bare `KeyError`, but the inconsistency is mildly surprising; consider running `IntTensor`'s `dtype` through the same `_NP_TO_SB`/`np.dtype(...).type` normalization for symmetry.

**3. Test inputs are slightly indirect.** `test_float_tensor_explicit_numpy_dtype` / `test_float_tensor_bad_dtype_raises` pass Python lists (`[1.0]`), which `FloatTensor` converts to an ndarray before the new branch runs — so they do exercise the fix, but the PR description headlines the `ndarray` case (`FloatTensor(arr, dtype=np.float32)`). Adding one explicit `np.array(...)` case (and a `np.dtype('float32')` case) would more directly cover the documented behavior. Minor.

### Verdict
No correctness bugs introduced by the diff itself. The fixes are correct and the mixed-precision / non-element / dtype paths are genuinely well covered. The one substantive gap is that the `BarcodeTensor` raw-array path inherits a pre-existing `Barcode` constructor bug, so bad barcode leaves still produce an `AttributeError` rather than the clean `TypeError` the PR promises — worth a small follow-up.

---

Issue #94, opened 2026-06-07
[friction/matrices] No way to read or set the storage as a numpy array on DistanceMatrixTensor /


bwehlin:
_Usability friction found by the masspcf API scan (2026-06-07). Not a correctness bug; see `friction-triage.md`._  **Priority: medium.**

### Scenario
A user has a stack (tensor) of N distance matrices and wants to populate or read them from a (N, n, n) numpy array, or get them all out as a dense (N, n, n) array.

### Current approach
```
No bulk API. You must loop and build each DistanceMatrix.from_dense(...) and assign cell-by-cell:
  t = mpcf.zeros((N,), dtype=mpcf.distmat64)
  for i in range(N):
      t[i] = DistanceMatrix.from_dense(arr[i])
and to read back:
  out = np.stack([t[i].to_dense() for i in range(N)])
The tensor itself has no from_dense / to_dense.
```

### Desired API
DistanceMatrixTensor.from_dense(ndarray_NxNxn) / .to_dense() (and the symmat equivalents) operating over the leading tensor axes, mirroring the scalar DistanceMatrix.from_dense/to_dense.

### Notes
The scalar matrix classes have from_dense/to_dense but the tensor variants do not, forcing per-element Python loops -- exactly the anti-pattern CLAUDE.md warns against.

Issue #93, opened 2026-06-07
[friction/matrices] No way to convert a PointCloudTensor (of equal-sized clouds) back to a single numpy array


bwehlin:
_Usability friction found by the masspcf API scan (2026-06-07). Not a correctness bug; see `friction-triage.md`._  **Priority: medium.**

### Scenario
Inverse of the above: after building/sampling a (m,) or (m,n) tensor of equally-sized (k,l) clouds, the user wants a dense (m,k,l) / (m,n,k,l) ndarray for plotting, sklearn, saving as .npy, etc.

### Current approach
```
np.asarray(point_cloud_tensor) raises 'TypeError: np.asarray() is not supported for PointCloudTensor.' There is no to_dense() on PointCloudTensor (unlike DistanceMatrix/SymmetricMatrix which have to_dense). Current workaround is a per-element loop:
  out = np.stack([np.asarray(pt[i]) for i in range(len(pt))])
which only works for 1-D tensors of equal-sized clouds and requires manual reshape for higher rank.
```

### Desired API
PointCloudTensor.to_dense() returning an ndarray of shape tensor_shape + cloud_shape when all clouds share a shape (else raise), and/or supporting np.asarray for the equal-shape case.

### Notes
DistanceMatrix and SymmetricMatrix both expose to_dense(); the point-cloud family lacks the symmetric capability.

PR #152, opened 2026-06-26
NumPy/dense export for point-cloud, matrix, and barcode tensors (#85, #93, #94)

### LINKED ISSUE #85: [friction/persistence] No way to get all bars out of a BarcodeTensor without a manual nested loop

### Scenario
After computing persistent homology over a batch, a user wants to export every barcode (e.g. to numpy / disk / plotting) at once.

### Current approach
```
np.asarray(bcs) raises TypeError ('np.asarray() is not supported for BarcodeTensor'); BarcodeTensor has no to_numpy(). The user must write nested loops over bcs.shape and call .to_numpy() on each Barcode element.
```

### Desired API
BarcodeTensor.to_numpy() returning an object ndarray of (n_i,2) arrays (or a flat list), or BarcodeTensor.tolist(); barcodes are ragged so a single dense array is impossible, but a documented batched accessor would remove the boilerplate.

### Notes
Barcode (singular) has to_numpy() and __array__; the tensor has neither.

bwehlin:
## Summary

Adds NumPy / dense export to the non-numeric tensor families, the inverse of the existing array/list constructors. Previously the only way to read these tensors out was a per-element loop, and `np.asarray()` raised `TypeError`.

- **PointCloudTensor (#93)** — `to_dense()` / `to_numpy()` / `__array__`. For a 0-d tensor returns the single `(n_points, dim)` cloud; otherwise stacks the equal-sized clouds into `(*tensor_shape, n_points, dim)`. `np.asarray(pc)` now materializes a real numeric array. Ragged clouds (or an empty leading axis) raise a clear `ValueError`.
- **DistanceMatrixTensor / SymmetricMatrixTensor (#94)** — `to_dense()` / `to_numpy()` / `__array__` returning `(*tensor_shape, n, n)`, plus a `from_dense` classmethod that aliases `from_numpy` so the API has the `from_dense`/`to_dense` pair. Both share a small `_matrix_tensor_to_dense` helper in `distance_matrix.py` (imported lazily from `symmetric_matrix.py`). Non-uniform matrix sizes / empty tensors raise `ValueError`.
- **BarcodeTensor (#85)** — `to_numpy()` returns an `object` ndarray (per-cell `(n_i, 2)` arrays) and `tolist()` the corresponding nested list; for a 0-d tensor both return the single `(n, 2)` array. Barcodes are ragged, so no `__array__` is added (the explicit methods are kept).

## Tests

New `test/python/test_nonnumeric_export.py` (21 cases): round-trips for point-cloud and matrix tensors, `np.asarray` parity and dtype argument, multi-dim leading shapes, `from_dense` alias, ragged/empty `ValueError` paths, 0-d single-cloud, and BarcodeTensor `to_numpy`/`tolist` (built from a list of `Barcode`s and via `compute_persistent_homology`).

Full suite: **1820 passed**, zero failures.

Docs (`docs/tensors.rst`, `docs/persistence.rst`) and `CHANGELOG.md` updated.

Closes #85, #93, #94.


bwehlin:
## 🔍 Adversarial review

### Summary

This PR adds NumPy/dense export (`to_dense`/`to_numpy`/`__array__`, plus `from_dense` aliases on the matrix tensors and `to_numpy`/`tolist` on `BarcodeTensor`) as the inverse of the existing constructors. I tried hard to break it and the implementation is **solid and correct**: the C-order iteration in `np.ndindex` matches `np.stack(...).reshape(tensor_shape + elem_shape)`, so multi-dim leading shapes land in the right place (and the `np.arange` round-trip tests actually pin this down). The empty-axis and ragged guards are reachable and tested for point clouds. No likely bugs found. A couple of minor gaps below.

### 🟢 Nits / minor

- **`_matrix_tensor_to_dense` non-uniform-size branch is untested.** `distance_matrix.py` raises `ValueError("...matrices have differing sizes...")` when `any(m.shape != (n, n))`, and the PR description explicitly claims "Non-uniform matrix sizes ... raise `ValueError`". But the new tests only cover the *empty-axis* path for matrices (`test_distance_matrix_tensor_empty_to_dense_raises`, `test_symmetric_matrix_tensor_empty_to_dense_raises`) — there is no test that actually builds a matrix tensor with differently-sized cells (e.g. assigning a 3×3 `DistanceMatrix` into one cell and a 4×4 into another) and asserts the raise. The point-cloud side *does* test this (`test_pointcloud_ragged_to_dense_raises`), so the matrix side is asymmetric. The branch looks correct; it's just not exercised, so a future regression there would pass CI silently.

- **`copy=` argument is ignored in all the new `__array__` overrides.** This mirrors the pre-existing `Barcode`/`DistanceMatrix.__array__` pattern and is in practice harmless here: `to_dense()` always returns a freshly `np.stack`-ed array that never aliases the C++ buffers, so even `copy=False` cannot produce a forbidden aliasing-without-copy situation and `np.array(..., copy=True)` already gets a fresh array. Worth a one-line acknowledgement rather than a change. (Consistency only.)

### 🟡 Edge case to confirm (low confidence)

- **`np.asarray(BarcodeTensor)` raising `TypeError`** (asserted by `test_barcode_tensor_no_array_protocol`) relies on NumPy *propagating* the base-class `Tensor.__array__` `TypeError` rather than falling back to the sequence protocol — and `Tensor` is iterable (`__iter__`/`__len__`). This is the long-standing contract for the other non-numeric tensors (PcfTensor etc.), so it almost certainly holds on the supported NumPy versions; flagging only because I couldn't execute it in this environment (no built package). If CI is green on the project's NumPy floor/ceiling, this is fully settled.

### Verdict

The change does what the description claims, the densification ordering is provably correct, and the happy-path + empty/ragged guards are well covered. The only substantive follow-up is adding a direct test for the non-uniform matrix-size `ValueError` to match the point-cloud ragged test. Looks good to merge.

---

PR #153, opened 2026-06-26
Reject NaN distance-matrix entries, not just negatives (#43)

### LINKED ISSUE #43: [matrices] DistanceMatrix nonnegativity check (`value < 0`) lets NaN bypass the documented "nonnegative"

### Summary
The validation almost certainly uses `value < 0`, which is False for NaN, so NaN slips through the nonnegativity guarantee documented in the class docstring ('symmetric, zero diagonal, nonnegative'). Low severity because it requires explicitly inserting NaN.

### Minimal reproduction
```python
import numpy as np
from masspcf.distance_matrix import DistanceMatrix

dm = DistanceMatrix(3)
dm[0, 1] = np.nan          # accepted silently
print('nan accepted, reads:', dm[0, 1], 'isnan:', np.isnan(dm[0, 1]))
assert np.isnan(dm[0, 1])  # NaN slips through the nonnegativity guard

# Contrast: negative values and -inf are rejected
for bad in (-1.0, -np.inf):
    try:
        dm[0, 2] = bad
        print(f'{bad} accepted (unexpected)')
    except ValueError as e:
        print(f'{bad} rejected:', e)

# +inf is also accepted (arguably valid distance, less clear-cut than NaN)
dm[1, 2] = np.inf
print('+inf accepted, reads:', dm[1, 2])
```

### Observed
Confirmed (exit 0, no crash). `dm[0,1] = np.nan` is accepted silently and reads back as nan; `np.isnan(dm[0,1])` is True. `dm[1,2] = np.inf` is also accepted. By contrast `dm[0,2] = -1.0` and `dm[0,2] = -np.inf` both raise ValueError "Distance matrix entries must be nonnegative". So NaN (and +inf) bypass the nonnegativity validation while negatives and -inf are rejected. The class docstring (line 47) declares "symmetric, zero diagonal, nonnegative" and `__setitem__` (line 134) states "entries must be nonnegative"; the C++ header (distance_matrix.hpp line 37) repeats "nonnegative entries". No existing test sanctions NaN/inf acceptance; tests only pin rejection of negatives (test_distance_matrix.py:62-65, test_pdist.py from_dense). Installed package (/opt/venv/lib/python3.10/site-packages/masspcf/distance_matrix.py) is byte-identical to the source tree, so this reflects current code.

### Expected
A NaN should be rejected (raise ValueError) the same way a negative value is, because NaN is not a valid nonnegative distance and the docstring/contract explicitly promises nonnegative entries. At minimum the guard should reject NaN; the +inf case is more defensible (an infinite distance can be meaningful in TDA, e.g. disconnected components) and could legitimately be allowed, but NaN should not slip through silently. The fix is to make the check reject anything that is not >= 0 rather than only what is < 0.

### Root-cause hint
include/mpcf/distance_matrix.hpp, DistanceMatrix::EntryProxy::operator= (line 63): the guard `if (value < T{})` is false for NaN (IEEE-754), so NaN bypasses the nonnegativity check; same `< T{}` pattern is duplicated in the DistanceMatrix(n, init) ctor (line 84) and in py_distance_matrix.hpp from_dense (line 83). Replacing `value < T{}` with `!(value >= T{})` would reject NaN and negatives while preserving current handling of finite values.

```python
# Bug 42: DistanceMatrix nonnegativity check (value < 0) lets NaN bypass the
#         documented "nonnegative" guarantee.
def test_distance_matrix_rejects_nan():
    """DistanceMatrix must reject NaN like it rejects negatives."""
    # BUG: the nonnegativity guard uses `value < 0`, which is False for NaN, so
    #      NaN slips through the documented "nonnegative entries" contract.
    # Expected: assigning NaN raises ValueError (NaN is not a valid nonnegative
    #           distance), the same way negatives and -inf are rejected.
    # Observed today: dm[0,1] = np.nan is accepted silently and reads back nan.

    dm = DistanceMatrix(3)
    # sanity: negatives already raise (pins the contract we extend to NaN).
    with pytest.raises(ValueError, match="nonnegative"):
        dm[0, 2] = -1.0

    with pytest.raises(ValueError, match="nonnegative"):
        dm[0, 1] = np.nan
```

</details>

bwehlin:
## Summary

`DistanceMatrix`'s nonnegativity guard used `value < T{}`, which is **false for NaN** (IEEE-754), so a NaN distance was stored silently despite the documented "nonnegative" contract. Via `from_dense` it surfaced even more confusingly as `Matrix must be symmetric` (because `NaN != NaN` failed the symmetry check after the nonneg check was bypassed).

## Fix

All three guard sites now use `!(x >= T{})`, which rejects both negatives and NaN while still accepting `+inf` (a meaningful infinite distance in TDA):
- `EntryProxy::operator=` (element assignment) and the `(n, init)` constructor in `include/sbear/distance_matrix.hpp`
- `from_dense` in `src/python/py_distance_matrix.hpp` (now checked before the symmetry check, so a NaN gives the correct "nonnegative" message)

## Tests

- Python `test/python/test_distance_matrix.py`: NaN rejected on element assignment and `from_dense` (matching "nonnegative"); negatives still rejected; `+inf` accepted; parametrized over float32/float64.
- C++ `test/test_tensor.cpp`: `DistanceMatrixNonnegTyped` (float/double) asserts NaN and negative throw `std::invalid_argument`, `+inf` is accepted and stored, and a NaN fill value in the `(n, init)` constructor throws.

Full C++ suite (444 tests) and Python suite (1805) pass.

Closes #43


bwehlin:
## 🔍 Adversarial review

**Summary.** This PR replaces the `value < T{}` nonnegativity guard with `!(value >= T{})` at three sites (`EntryProxy::operator=`, the `(n, init)` constructor, and `from_dense`), so NaN is now rejected alongside negatives while `+inf` stays allowed. The core logic is correct (`!(NaN >= 0)` is `true`, `!(+inf >= 0)` is `false`, `-0.0` and integers unaffected), only float instantiations exist, and the tests are properly parametrized over f32/f64. The change is sound; the items below are about behavior the description doesn't mention rather than defects in the guard itself.

### 🟡 Edge cases to confirm

**1. The same guard now fires inside the `pdist` compute path, not just user setters.** Both the result writers and the integration task assign computed distances through `EntryProxy::operator=`:
- `include/sbear/cuda/cuda_result_writer.hpp:91` — `m_mat(iGlobal, jGlobal) = hostBlock[...]`
- `include/sbear/algorithms/functional/matrix_integrate.hpp:87` — `m_out(i, j) = val;` (inside `compute_row`, run on taskflow workers)

Pre-PR, a NaN distance produced by integration (e.g. `inf - inf` / `0 * inf` in an Lp combine) was stored silently; post-PR it throws `std::invalid_argument` from inside a worker task. Negatives already threw here, so throwing-from-worker isn't new — but the NaN trigger is, and this is an unannounced behavior change in the hot path. Worth confirming (a) that this is intended hardening and (b) that a `std::invalid_argument` raised inside a taskflow task surfaces cleanly to the Python caller rather than being swallowed/aborting. No test covers the "pdist produces NaN" path; the tests only exercise the user-facing setters and `from_dense`.

**2. `from_dense`'s "correct nonnegative message" only holds for upper-triangle / symmetric NaN.** The nonneg check inspects only `r(i, j)` for `j > i` (`src/python/py_distance_matrix.hpp:67`). A NaN present *only* in the lower triangle passes the nonneg check and then trips `r(i,j) != r(j,i)` → still `"Matrix must be symmetric"`. The added `test_reject_nan_from_dense` only sets a symmetric NaN (`arr[0,1] = arr[1,0] = nan`), so this asymmetry is untested. (Pre-existing structure; the PR's message-correctness claim is just partial.)

### 🟢 Nits

- **Diagonal NaN now reports "nonnegative" instead of "Diagonal must be zero".** For `dm[i, i] = nan`, the new nonneg check runs before the `!m_ptr` diagonal branch, so the message changes from `"Diagonal entries of a distance matrix must be zero"` to `"...nonnegative"` (`include/sbear/distance_matrix.hpp`). Arguably more correct, but it's a user-visible message change worth a one-line note. Note `from_dense` keeps reporting `"Diagonal entries must be zero"` for a diagonal NaN (the diagonal `!= T{}` check runs first there), so the two paths now disagree on the message for a diagonal NaN.
- Consider asserting the rejection happens *before* any partial mutation in `from_dense` (it does — `dm` is built only after validation — but no test pins this).

**Verdict:** The guard change is correct and well-tested for the user-facing surface. The one thing I'd want confirmed before merge is the new NaN-throws-from-`pdist`-worker behavior (finding 1), since that's a real semantic change the description doesn't call out.

---

PR #154, opened 2026-06-26
Propagate RuntimeWarning from ** under -W error instead of SystemError (#12)

### LINKED ISSUE #12: [pcf] Pcf.__pow__ (and tensor __pow__) raise SystemError instead of propagating RuntimeWarning when

### Summary
The __pow__ docstring promises 'A RuntimeWarning is emitted if the result contains NaN or infinity'. test_pow_negative_base_fractional_exp_warns relies on this. Running the existing tests under the very common `-W error` / `filterwarnings=error` configuration turns this documented warning into a SystemError with a meaningless message, because the C++ binding (py_tensor.hpp PyErr_WarnEx) ignores the -1 return code and proceeds to build a return value with a pending Python exception.

### Minimal reproduction
```python
import warnings, numpy as np, masspcf as mpcf
from masspcf.functional import Pcf
f = Pcf(np.array([[0.0, -4.0], [1.0, 9.0]], dtype=np.float64))
try:
    with warnings.catch_warnings():
        warnings.simplefilter('error')   # same as python -W error / pytest filterwarnings=error
        g = f ** 0.5
    print('result', g.to_numpy().tolist())
except RuntimeWarning as e:
    print('clean (expected):', repr(e))
except SystemError as e:
    print('BUG SystemError:', str(e)[:200])
# Output: BUG SystemError: <built-in method __pow__ ...> returned a result with an exception set
```

### Observed
Under warnings.simplefilter('error') (equivalent to python -W error / pytest filterwarnings=error), `Pcf(...) ** 0.5` for a PCF containing a negative value raises `SystemError: <built-in method __pow__ of pybind11_builtins...> returned a result with an exception set` rather than the documented RuntimeWarning. The same fault occurs on the tensor path: `mpcf.FloatTensor(np.array([-4.0, 9.0])) ** 0.5` also raises SystemError under -W error. Under default/'always' warning settings the behavior is correct: a RuntimeWarning ('invalid or infinite value encountered in pcf pow') is emitted and nan is returned (e.g. [[0.0, nan], [1.0, 3.0]]).

### Expected
When PyErr_WarnEx for the RuntimeWarning is escalated to an error, the binding should propagate that as a clean RuntimeWarning/RuntimeError exception (no value returned with a pending exception). The docstring (both Pcf.__pow__ in py_pcf.cpp and the base in masspcf/_tensor_base.py:884) promises 'A RuntimeWarning is emitted if the result contains NaN or infinity', and -W error / filterwarnings=error are supported, common configurations under which that warning must surface as a normal exception, not an interpreter-level SystemError with a meaningless message.

### Root-cause hint
src/python/functional/py_pcf.cpp:151-163 __pow__ lambda (and identically src/python/py_tensor.hpp:453-469): the code calls PyErr_WarnEx(PyExc_RuntimeWarning, ...) but ignores its return value; when warnings are filtered to errors PyErr_WarnEx returns -1 and sets a pending Python exception, yet the lambda proceeds to `return result`, so pybind11 builds a return value with an exception already set, which CPython reports as 'returned a result with an exception set' (SystemError). Fix: check the -1 return and `throw py::error_already_set();` (do not return a value).

In `test/python/test_bugscan_pcf.py` -- run with `cd test && python -m pytest python/test_bugscan_pcf.py`

```python
# Bug 11: __pow__ raises a meaningless SystemError instead of propagating the
#         RuntimeWarning when warnings are configured as errors.
def test_pcf_pow_under_warnings_as_errors_raises_clean_runtimewarning():
    """Pcf.__pow__ must surface the documented RuntimeWarning cleanly under -W error.

    Under ``warnings.simplefilter('error')`` (== ``python -W error`` /
    ``filterwarnings=error``) a fractional power of a negative-valued PCF should
    raise the documented ``RuntimeWarning`` as a normal exception. Instead the
    binding ignores ``PyErr_WarnEx``'s -1 return and builds a return value with a
    pending exception, which CPython reports as a meaningless ``SystemError``.
    """
    # BUG: py_pcf.cpp __pow__ ignores PyErr_WarnEx's -1 return code; under
    #      warnings-as-errors it returns a value with a pending exception set.
    # Expected: a clean RuntimeWarning (escalated to an exception) is raised.
    # Observed today: SystemError "... returned a result with an exception set".
    f = Pcf(np.array([[0.0, -4.0], [1.0, 9.0]], dtype=np.float64))
    with warnings.catch_warnings():
        warnings.simplefilter("error")
        # The documented contract: the RuntimeWarning becomes a clean exception.
        # Today this raises SystemError instead, which pytest.raises(RuntimeWarning)
        # does NOT catch -> the SystemError propagates and fails the test cleanly.
        with pytest.raises(RuntimeWarning):
            f ** 0.5
```

</details>

bwehlin:
## Summary

`Pcf.__pow__` and the tensor `__pow__` emit a `RuntimeWarning` when a result contains NaN/inf, via `PyErr_WarnEx`, but ignored its return value. When warnings are escalated to errors (`python -W error` or `warnings.simplefilter("error")`), `PyErr_WarnEx` returns `-1` and leaves a Python exception pending — the bindings then returned a value anyway, so pybind11 raised a confusing `SystemError` instead of the documented `RuntimeWarning`.

## Fix

Both binding sites (`src/python/functional/py_pcf.cpp`, `src/python/py_tensor.hpp`) now check the `PyErr_WarnEx` return code and `throw py::error_already_set()` when it is non-zero, so the escalated warning propagates as the expected `RuntimeWarning`. Default behavior (emit the warning and continue) is unchanged, and a finite result still emits no warning.

## Tests

`test/python/test_bugscan_pcf.py`: under `simplefilter("error")`, both `Pcf ** 0.5` and `FloatTensor ** 0.5` (negative base) raise `RuntimeWarning` (not `SystemError`); the warning is still emitted by default; an all-positive base emits no warning even under `-W error`.

C++ suite (444) and Python suite (1803) pass.

Closes #12


bwehlin:
## 🔍 Adversarial review

### Summary
This PR makes `Pcf.__pow__` and tensor `__pow__` check the return code of `PyErr_WarnEx` and `throw py::error_already_set()` when the `RuntimeWarning` has been escalated to an exception (`-W error` / `simplefilter("error")`), so the documented `RuntimeWarning` propagates instead of pybind11's confusing `SystemError`. The change is small, correct, and well-tested — I could not find a substantive defect. Verdict: **looks good to merge.**

### What I checked (and why it holds up)
- **Coverage of warning sites is complete.** `PyErr_WarnEx` appears in exactly two places in `src/python` (`functional/py_pcf.cpp:141` and `py_tensor.hpp:449`), and both are fixed identically. There is no `__rpow__`, and `__ipow__` (`py_tensor.hpp:455`) never emitted a warning in the first place, so nothing was missed by this patch.
- **Exception/RAII safety is fine.** In both sites `result` is a stack local destroyed by RAII when `throw` unwinds; `error_already_set()` captures the already-pending Python exception (the escalated `RuntimeWarning`) and re-raises that same object. No leak, no double-set.
- **Tests genuinely discriminate the fix from the bug.** `SystemError` and `RuntimeWarning` are disjoint in the class hierarchy (`SystemError → Exception`; `RuntimeWarning → Warning → Exception`), so `pytest.raises(RuntimeWarning)` in `test_pcf_pow_propagates_warning_as_error` / `test_tensor_pow_propagates_warning_as_error` would have **failed** against the old `SystemError`-raising code. These are real regression guards, not no-ops.
- **Both positive and negative paths are exercised.** The NaN-producing inputs are correct (`(-2.0)**0.5 → NaN` for both Pcf and `FloatTensor`), `test_pcf_pow_warning_still_emitted_by_default` confirms unchanged default behavior, and `test_pcf_pow_no_warning_for_finite_result` (base 4/9, exp 0.5) confirms no false-positive warning under `-W error`.

### Findings

#### 🟢 Nits / out-of-scope observations
- **`__ipow__` is inconsistent with `__pow__` (pre-existing, not introduced here).** In-place power (`py_tensor.hpp:455`) calls `sb::ipow` and emits **no** warning at all, even when it produces NaN/inf. After this PR, `t ** 0.5` warns/raises but `t **= 0.5` stays silent. Out of scope for #12, but worth a follow-up issue for consistency.
- **Tests don't assert the warning *message*.** `pytest.raises(RuntimeWarning)` matches any `RuntimeWarning`, so the "invalid or infinite value encountered in pow" text isn't pinned. Low value given the class-discrimination above; optionally add `match="invalid or infinite"` to lock the message.
- **No direct test that the result of a non-escalated warning is still returned correctly** for the tensor `__pow__` (the Pcf default-emit test covers the Pcf side). Minor, since the code path is shared logic.

No likely bugs and no risky edge cases found — the change is narrow and the fix matches the bug report precisely.

---

Issue #11, opened 2026-06-07
[pcf] Pcf construction validates t0 on the unsorted input row 0, then std::sort can move a negative time


bwehlin:
_Found by an automated masspcf API bug scan (2026-06-07); independently reproduced and verified. **Severity: high.**_

### Summary
Two documented guarantees are violated simultaneously: (1) the class docstring states t0 must be 0 and an InvalidArgument is raised otherwise; (2) __call__ docstring states 'Raises ValueError if any time is negative (PCFs are defined on [0, inf))'. Both fail because the validation runs on the unsorted input and the eval guard checks t0 rather than 0. This is silent: no error, just a malformed PCF that returns values at negative times.

### Minimal reproduction
```python
import numpy as np
from masspcf.functional import Pcf
f = Pcf(np.array([[0.0, 1.0], [2.0, 2.0], [-1.0, 99.0]], dtype=np.float64))
print('stored:', f.to_numpy().tolist())   # [[-1.0,99.0],[0.0,1.0],[2.0,2.0]] -> t0=-1.0
print('f(-0.5)   =', f(-0.5))              # 99.0  (no error)
print('f([-0.5]) =', f(np.array([-0.5])).tolist())  # [99.0]  (no error)
# Contrast: a well-formed PCF correctly raises
g = Pcf(np.array([[0.0, 1.0], [2.0, 2.0]], dtype=np.float64))
try:
    g(-0.5)
except ValueError as e:
    print('well-formed g(-0.5) raises:', e)  # 'Cannot evaluate PCF before time 0.'
```

### Observed
Pcf(np.array([[0.0,1.0],[2.0,2.0],[-1.0,99.0]])) constructs successfully and stores [[-1.0,99.0],[0.0,1.0],[2.0,2.0]] with t0=-1.0. The class docstring guarantees t0=0 (and that an InvalidArgument is raised otherwise), but no error is raised. Then f(-0.5) returns 99.0 and f([-0.5]) returns [99.0] instead of raising ValueError; the __call__ docstring promises ValueError for negative times. The malformed PCF only raises for t < -1.0 (e.g. f(-1.5)), confirming the eval guard compares against front().t (=t0), not 0. By contrast, a well-formed PCF (rows already start at t=0) correctly raises 'Cannot evaluate PCF before time 0.' for both scalar -0.5 and array [-0.5], and an input whose ROW 0 is nonzero is correctly rejected with 'The first breakpoint must have time t=0'.

### Expected
[...]

PR #155, opened 2026-06-26
Pcf rejects duplicate breakpoint times and empty input (#56)

### LINKED ISSUE #56: [friction/construction] Pcf silently sorts non-monotonic / duplicate breakpoints instead of raising as documented

### Scenario
User constructs a Pcf from rows whose times are not strictly increasing (e.g. unsorted, or with duplicate times).

### Current approach
```
f = mpcf.Pcf(np.array([[0.0,100.0],[5.0,500.0],[2.0,200.0]], dtype=np.float64))
f.to_numpy()  # -> [[0,100],[2,200],[5,500]]  (silently sorted by time)
mpcf.Pcf(np.array([[0.0,5.0],[1.0,6.0],[1.0,7.0]]))  # duplicate t=1 silently accepted
Also mpcf.Pcf(np.zeros((0,2))) reports .size == 1 (a single [0,0] breakpoint) for empty input.
```

### Desired API
Per the Pcf docstring ('t0 < t1 < ... < t_{n-1}'), construction from non-strictly-increasing times should raise ValueError (matching the existing t0!=0 check), rather than silently reordering rows or accepting duplicates. Empty (0,2) input should give size 0 or raise, not size 1.

### Notes
Whole rows are sorted together so (time,value) pairs stay associated, so this is a documented-guarantee mismatch rather than corruption; still surprising and contradicts the docstring.

bwehlin:
## Summary

Completing the stricter `Pcf` validation from #11/#124:

- **Duplicate times** — the order check used `std::is_sorted` with a `<` comparator, which treats equal consecutive times as "sorted", so a `Pcf` with duplicate breakpoint times (e.g. two rows at `t=0`) was silently accepted, contradicting the documented strictly-increasing contract (`t_0 < t_1 < ...`).
- **Empty input** — `Pcf(np.zeros((0, 2)))` silently fabricated a single `(0, 0)` breakpoint instead of reporting that a PCF needs at least one breakpoint.

## Fix (`src/python/pypcf_support.hpp::construct_pcf`)

- After the non-decreasing-order check, reject any duplicate time via `std::adjacent_find` (clear `ValueError`: "strictly increasing times").
- Reject an empty input array with a `ValueError` (use `Pcf()` for the all-zero constant PCF — its default constructor is unchanged).

`from_serial_content` is a separate loose deserialization path (it does not go through `construct_pcf`) and is unaffected — it can still produce duplicate-time PCFs from serialized data.

Docstring (`stablebear/functional/pcf.py`) and `docs/concepts.rst` updated to say "strictly increasing" and note the empty-array rejection.

## Tests

- `test/python/test_bugscan_pcf.py`: duplicate times (interior and leading) and empty `(0, 2)` input raise `ValueError`; a strictly-increasing PCF still constructs.
- Updated two pre-existing tests that fed degenerate data through the now-strict `Pcf(array)` constructor: `test_from_serial_content.py::test_fsc_multidim` (slices with all-`t=0` rows) now compares the produced PCF's breakpoints to the source slice directly; `test_numpy.py::test_np_to_pcf_has_correct_type` (which used `np.zeros((2,2))` purely for dtype inference) now uses valid strictly-increasing times.

Full Python suite (1803) and C++ suite (442) pass.

Closes #56

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Breaking changes
 
 * **`Pcf` no longer sorts breakpoints; unsorted, negative, and offset times are rejected** — `Pcf` construction used to silently sort the supplied `(time, value)` rows and checked the `t=0` requirement on the unsorted input row 0, so out-of-order breakpoints were accepted and a negative time on a later row could be reordered to the front into a malformed PCF that then evaluated at negative times. Breakpoints must now be supplied in non-decreasing time order (out-of-order input raises `ValueError` instead of being silently sorted), and the first time must be 0 — so any input whose first time is not 0, in particular any negative time, raises `ValueError`. `Pcf` evaluation also rejects `t < 0` directly (it previously compared against the first breakpoint rather than 0). ([#11](https://github.com/kthtda/stablebear/issues/11))
+* **`Pcf` requires strictly increasing breakpoint times and rejects empty input** — completing the `Pcf` validation above: a duplicate breakpoint time previously slipped past the `non-decreasing` check despite the documented strictly-increasing contract, and is now rejected with `ValueError`. Constructing a `Pcf` from an empty `(0, 2)` array (which used to silently fabricate a single `(0, 0)` breakpoint) now raises `ValueError`; use `Pcf()` for the all-zero constant PCF. ([#56](https://github.com/kthtda/stablebear/issues/56))
 
 ### Bug fixes
 
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -27,7 +27,7 @@ In stablebear, a PCF is represented as an :math:`n \times 2` array of ``(time, v
 
 The value at each breakpoint is the value on the interval starting at that time and continuing until the next breakpoint (or until the end of the function's domain).
 
-The breakpoints must be listed in non-decreasing order of time, and the first time must be ``0`` (PCFs are defined on :math:`[0, \infty)`). Breakpoints given out of order, a first time other than ``0``, or any negative time raise a ``ValueError``.
+The breakpoint times must be strictly increasing (:math:`t_0 < t_1 < \cdots`), and the first time must be ``0`` (PCFs are defined on :math:`[0, \infty)`). Out-of-order or duplicate times, a first time other than ``0``, any negative time, or an empty array raise a ``ValueError``.
 
 Why PCFs?
 ---------
--- a/src/python/pypcf_support.hpp
+++ b/src/python/pypcf_support.hpp
@@ -41,8 +41,11 @@ namespace sb
       py::buffer_info buf = arr.request();
       if (buf.size == 0)
       {
-        points.emplace_back(static_cast<Tt>(0), static_cast<Tv>(0));
-        return sb::Pcf<Tt, Tv>(std::move(points));
+        // A PCF must have at least one breakpoint at t=0; an empty input array
+        // cannot represent one. (Use Pcf() for the all-zero constant PCF.)
+        throw std::invalid_argument(
+          "Cannot construct a Pcf from an empty array; a PCF must have at "
+          "least one breakpoint at t=0.");
       }
 
       if (buf.ndim != 2)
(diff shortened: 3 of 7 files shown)


bwehlin:
## 🔍 Adversarial review

### Summary
The PR tightens `Pcf(array)` construction to reject duplicate breakpoint times and empty input, completing the strictly-increasing contract from #11/#124. The core change is **correct and well-targeted**: I tried to break it and could not find a correctness bug in the validation logic itself or the tests. One real behavioral edge case (`astype` round-trips) is worth confirming, plus a couple of nits.

### 🟡 Edge cases to confirm

**`Pcf.astype()` now fails for legitimately-constructed duplicate-time PCFs** — `stablebear/functional/pcf.py:129`
```python
def astype(self, dtype):
    ...
    return Pcf(self.to_numpy().astype(np_dtype))   # round-trips through strict construct_pcf
```
`astype` rebuilds the PCF by feeding its own points back through the now-strict `Pcf(array)` constructor. The PR description itself notes that `from_serial_content` is a separate loose path that "can still produce duplicate-time PCFs." But such a PCF can no longer be cast: `from_serial_content(...)[i].astype(pcf32)` will now raise `ValueError("strictly increasing ...")`, even though nothing in the data changed and `astype` is documented as a pure dtype cast. This is an untested, user-visible behavior change. (A latent variant of this existed for the #11 sort/order rule too, but duplicate-time is the realistic trigger since `from_serial_content` preserves source order.) Worth either confirming it's acceptable, routing `astype` through the loose path, or at least adding a test pinning the behavior.

### 🟢 Nits

- **New rules tested only on float64.** `construct_pcf` is a template instantiated for `i32`/`i64` as well (`Pcf_i32_i32`, `Pcf_i64_i64`). The duplicate-time and empty-array tests in `test/python/test_bugscan_pcf.py` only cover `np.float64`. The existing suite already tests an int path for the `t=0` rule (`test_intpcf_negative_first_time_rejected`), so an int duplicate-time case would match that precedent. Low risk since the logic is dtype-agnostic.

- **Duplicate-time `==` is exact float equality** (`pypcf_support.hpp`, `adjacent_find` with `a.t == b.t`). This is correct here — duplicates come verbatim from the input array, not from computation, so there's no tie-tolerance concern — but it's worth a one-line note that "duplicate" means bit-exact, in case callers expect near-duplicate coalescing.

### Things I checked that are fine (no action needed)
- **Exception type:** the new throws are `std::invalid_argument`, which pybind11 maps to `ValueError`; confirmed against the pre-existing `test_unsorted_input_rejected`/`test_negative_first_time_rejected` tests that already rely on this mapping for the same function. All new `pytest.raises(ValueError, match=...)` strings ("strictly increasing", "empty") match the thrown messages.
- **Check ordering:** the duplicate check is placed after the `is_sorted` (non-decreasing) check and before the `front().t != 0` check. For a non-decreasing sequence, duplicates are always adjacent, so `adjacent_find` is both sufficient and correct. The leading-duplicate case `[[0,5],[0,6]]` correctly hits the duplicate check (not the `t=0` check) — verified.
- **Modified `test_fsc_multidim`:** comparing `np.asarray(X[i,j])` to `content[start:stop]` is sound — `Pcf(std::vector&&)` and `from_serial_content` both store points verbatim (no sort/dedup/coalesce), and `to_numpy` returns them in the same Nx2 layout, so the round-trip is exact.
- **No collateral breakage:** swept the test suite, `stablebear/`, and `docs/` for `Pcf(np.zeros((N,2)))` with N≥2 (would now fail as all-`t=0` duplicates) — the only one was the `test_numpy.py` case the PR already fixed. `Pcf(np.zeros((1,2)))` usages (single `t=0` breakpoint) remain valid.
- `test_pcf_strictly_increasing_still_constructs` is a real positive control (`f.size == 3`), so the new rejections aren't over-broad.

**Verdict:** Solid, faithful to its description, and the tests actually exercise the claimed fixes. The only substantive thing to resolve is whether breaking `astype()` on duplicate-time PCFs (producible via `from_serial_content`) is intended.


---

PR #156, opened 2026-06-26
allclose: handle infinities, shape mismatch, and int/bool/ndarray (#51, #52, #100)

### LINKED ISSUE #51: [interop] allclose breaks reflexivity for +inf and gives false positives for opposite-sign infinities

### Summary
NumPy-semantics divergence on a documented 'element-wise equal within tolerance' helper; allclose(x, x) being False is surprising and breaks reflexivity for inf-containing data.

### Minimal reproduction
```python
x = mpcf.FloatTensor(np.array([1.0, np.inf, 3.0]))
print('masspcf allclose(x,x):', mpcf.allclose(x, x))            # False
print('numpy   allclose(same):', np.allclose([1.,np.inf,3.],[1.,np.inf,3.]))  # True
# false positive on opposite-sign infinities:
xp = mpcf.FloatTensor(np.array([np.inf]))
xm = mpcf.FloatTensor(np.array([-np.inf]))
print('masspcf allclose([inf],[-inf]):', mpcf.allclose(xp, xm))  # True (!)
print('numpy   allclose([inf],[-inf]):', np.allclose([np.inf],[-np.inf]))  # False
assert mpcf.allclose(x, x)  # AssertionError
```

### Observed
mpcf.allclose(x, x) returns False whenever x contains +inf (or -inf): |inf-inf| = nan, and nan <= atol+rtol*|b| is False, so reflexivity is broken. Worse, mpcf.allclose([inf],[-inf]) returns True (a silent false positive): |inf-(-inf)| = inf and atol+rtol*|b| = inf, so inf <= inf is True. Verified by hand and by running the repro from /tmp (exit code 1 due to the reflexivity assert failing). NumPy returns True for matching infinities and False for opposite-sign infinities.

### Expected
Matching the numpy.allclose semantics the implementation explicitly claims to follow: same-sign infinities at the same position should compare equal (allclose(x,x) True for any inf-containing, non-NaN tensor), and opposite-sign infinities should compare not-close (False). NumPy achieves this by comparing infinite elements with == rather than via the |a-b|<=atol+rtol*|b| formula.

### Root-cause hint
include/mpcf/tensor.tpp:allclose (lines 1676-1688 for Tensor, 1692-1706 for matrices) applies the bare formula std::abs(aval-bval) <= atol + rtol*std::abs(bval) with no inf special-case, despite the line 1674 comment "Follows the numpy convention"; inf-inf=nan makes equal infinities compare False (reflexivity break) and inf<=inf makes opposite-sign infinities compare True (false positive). Fix: special-case infinite elements with equality (a==b) as numpy's isclose does, in both the Tensor and is_compressed_matrix_v overloads.

```python
# Bug 51: allclose breaks reflexivity for +inf and gives false positives for
#         opposite-sign infinities.
def test_allclose_infinity_matches_numpy():
    """allclose must treat matching infinities as close and opposite-sign as not."""
    # BUG: allclose(x,x) is False when x has +inf; allclose([inf],[-inf]) is True
    # Expected: NumPy semantics -- matching infinities compare close (reflexivity
    #           holds), opposite-sign infinities compare not-close.
    # Observed today: allclose(x,x)->False (|inf-inf|=nan); allclose([inf],[-inf])->True.
    x = mpcf.FloatTensor(np.array([1.0, np.inf, 3.0]))
    # reflexivity must hold for inf-containing tensors (numpy: True)
    assert bool(mpcf.allclose(x, x)) == bool(
        np.allclose([1.0, np.inf, 3.0], [1.0, np.inf, 3.0]))

    xp = mpcf.FloatTensor(np.array([np.inf]))
    xm = mpcf.FloatTensor(np.array([-np.inf]))
    assert bool(mpcf.allclose(xp, xm)) == bool(
        np.allclose([np.inf], [-np.inf]))  # numpy: False
```

</details>

### LINKED ISSUE #52: [interop] allclose silently returns False on shape-mismatched tensors instead of broadcasting (when

### Summary
Silent wrong/misleading result: a shape bug in user code is masked as a numerical 'not close' answer instead of being surfaced.

### Minimal reproduction
```python
a = mpcf.FloatTensor(np.array([1.0,2.0,3.0]))      # (3,)
e = mpcf.FloatTensor(np.array([[1.0,2.0,3.0]]))    # (1,3), broadcast-equal
b = mpcf.FloatTensor(np.array([1.0,2.0]))          # (2,), incompatible
print('allclose (3,) vs (1,3):', mpcf.allclose(a, e), '| numpy:', np.allclose([1.,2.,3.], [[1.,2.,3.]]))
# -> masspcf: False | numpy: True
print('allclose (3,) vs (2,):', mpcf.allclose(a, b))
# -> masspcf: False ; numpy raises ValueError (operands could not be broadcast)
# Library DOES broadcast elsewhere:
print((a+e).shape)                 # Shape(1, 3)  -> arithmetic broadcasts
print(np.asarray(a==e).tolist())   # [[True, True, True]] -> comparison broadcasts
# and raises on truly incompatible shapes:
try:
    a + b
except ValueError as ex:
    print('a+b raised:', ex)       # Shapes are not broadcast-compatible: (3) and (2)
```

### Observed
mpcf.allclose returns False for any shape mismatch. For (3,) vs (1,3) with broadcast-equal values it returns False; for (3,) vs (2,) (genuinely incompatible) it also returns False. Root cause: mpcf::allclose in include/mpcf/tensor.tpp:1678 does `if (a.shape() != b.shape()) return false;` before any element comparison. This is inconsistent with the library's own broadcasting semantics: tensor arithmetic (a+e) and comparison (a==e) DO broadcast (3,) vs (1,3) to (1,3) with elementwise True, and tensor arithmetic DOES raise ValueError("Shapes are not broadcast-compatible") on (3,) vs (2,). The C++ source even comments "Follows the numpy convention" for the tolerance formula. Neither the docstring nor docs/.rst claim shape mismatch yields False, and no test pins this behavior (test/python/test_allclose.py only covers same-shape and type-mismatch cases).

### Expected
Consistent with NumPy and with the library's own broadcasting (used in +, ==, etc.): broadcast when shapes are compatible -- so (3,) vs (1,3) of equal values should return True -- and raise ValueError on truly incompatible shapes like (3,) vs (2,), rather than silently collapsing both to a misleading False that a user reads as 'values differ'.

### Root-cause hint
include/mpcf/tensor.tpp: mpcf::allclose(Tensor,Tensor) -- line 1678 short-circuits `if (a.shape() != b.shape()) return false;` instead of using the same broadcast machinery that tensor operators use (which broadcasts compatible shapes and raises on incompatible ones). The Python wrapper masspcf/comparison.py:allclose just forwards to a._data.allclose, so the fix belongs in C++.

```python
# Bug 52: allclose silently returns False on shape mismatch instead of
#         broadcasting (when compatible) or raising (when incompatible).
def test_allclose_broadcasts_or_raises_on_shape_mismatch():
    """allclose must broadcast compatible shapes and raise on incompatible ones."""
    # BUG: allclose returns False for ANY shape mismatch (short-circuit before
    #      comparing), inconsistent with the library's own broadcasting in + / ==
    # Expected: broadcast-compatible shapes -> compare elementwise (so (3,) vs
    #           (1,3) of equal values is True); truly incompatible -> ValueError.
    # Observed today: both cases silently return False.
    a = mpcf.FloatTensor(np.array([1.0, 2.0, 3.0]))       # (3,)
    e = mpcf.FloatTensor(np.array([[1.0, 2.0, 3.0]]))     # (1,3) broadcast-equal
    b = mpcf.FloatTensor(np.array([1.0, 2.0]))            # (2,) incompatible

    # broadcast-compatible, equal values -> True
    assert bool(mpcf.allclose(a, e)) is True

    # truly incompatible -> ValueError, not a silent False
        mpcf.allclose(a, b)
```

</details>

### LINKED ISSUE #100: [friction/interop] allclose only accepts FloatTensor/DistanceMatrix/SymmetricMatrix; IntTensor and BoolTensor are

### Scenario
A user wants to tolerance-compare two integer tensors, or compare a FloatTensor against a numpy array, with one helper.

### Current approach
```
mpcf.allclose(IntTensor, IntTensor), allclose(BoolTensor, BoolTensor), and allclose(FloatTensor, ndarray) all raise 'TypeError: allclose requires two objects of the same supported type (FloatTensor, DistanceMatrix, or SymmetricMatrix)'. Workaround: np.allclose(np.asarray(a), np.asarray(b)).
```

### Desired API
allclose should accept any numeric tensor (IntTensor/BoolTensor), accept an ndarray on either side (coercing like NumericTensor.__eq__ does), and broadcast like np.allclose -- so allclose(float_tensor, np_array) just works.

### Notes
Matches docstring as written (FloatTensor only), so reported as friction not a bug, but the float-vs-ndarray rejection is the surprising part.

bwehlin:
## Summary

Three related `allclose` fixes — two in the C++ core (so the C++ library is correct), one in the Python wrapper.

- **#51 (inf)** — the bare formula `|a - b| <= atol + rtol*|b|` yields `NaN` for `|inf - inf|`, so `allclose(x, x)` returned `False` when `x` contained `inf` (reflexivity broken), and opposite-sign infinities compared *close*.
- **#52 (shape)** — `allclose` on differently-shaped tensors silently returned `False`, hiding real shape bugs, instead of broadcasting or raising.
- **#100 (types)** — only `FloatTensor`/`DistanceMatrix`/`SymmetricMatrix` were accepted; `IntTensor`, `BoolTensor`, and NumPy arrays were rejected.

## Fix

**C++ (`include/sbear/tensor.tpp`)**
- New `scalar_allclose(a, b, atol, rtol)`: if either operand is infinite, compare with `==` (same-sign inf → close; finite-vs-inf / opposite-sign → not close); otherwise the numpy formula. `NaN` is never close (`equal_nan=False`). Used by both the `Tensor` and compressed-matrix overloads.
- The `Tensor` overload now `broadcast_shapes` + `broadcast_to`s the operands (raising `std::invalid_argument` → `ValueError` on incompatible shapes) instead of returning `false` on shape mismatch.

**Python (`stablebear/comparison.py`)**
- Numeric/boolean tensors and ndarrays are coerced to `float64` and routed through the fixed C++ `Tensor.allclose` (so they get the inf + broadcast behavior, and differing precisions like float32-vs-float64 now work).
- `DistanceMatrix`/`SymmetricMatrix` keep their compact C++ comparison for matching precision and fall back to `np.allclose` on the dense forms for differing precision.
- At least one operand must be a stablebear tensor/matrix (so `allclose(42, 42)` still raises), and a matrix compared against a different type still raises `TypeError`.

## Tests
- C++ `test/test_tensor.cpp`: incompatible shapes throw, broadcast-compatible shapes work, and an `InfinityComparison` case (reflexive +inf, +inf≠-inf, inf≠finite). The old `ShapeMismatch` (expected `false`) test is replaced.
- Python `test/python/test_allclose.py`: infinity cases (incl. matching `np.allclose`), broadcast + incompatible-shape, and interop (int/bool/int-vs-float/float32-vs-float64/ndarray); the two error tests updated to the new messages.

Full C++ suite (446) and Python suite (1811) pass.

Closes #51
Closes #52
Closes #100


bwehlin:
## 🔍 Adversarial review

### Summary

The PR does what it claims. `scalar_allclose` correctly fixes the `|inf - inf| = NaN` reflexivity bug (#51) for both the dense `Tensor` overload and the compressed-matrix overload; the `Tensor` overload now broadcasts via `broadcast_shapes`/`broadcast_to` and raises `std::invalid_argument` → `ValueError` on incompatible shapes (#52); and the Python wrapper routes `IntTensor`/`BoolTensor`/ndarray through the fixed C++ path as `float64` (#100). I verified the broadcast view indexing (stride-0 dims through `index_to_data_index`), `walk`'s short-circuit and empty/rank-0 handling, the pybind `atol`/`rtol` kwarg mapping, and that `NumericTensor`/`BoolTensor`/matrix `__array__` all support the `np.asarray(..., dtype=float64)` coercions used. No correctness bugs found — the findings below are test-quality and consistency nits.

### Findings

#### 🟡 Edge cases to confirm

- **The compressed-matrix overload still silently returns `false` on a size/storage mismatch** (`tensor.tpp`, unchanged `if (a.storage_count() != b.storage_count()) return false;`). Issue #52 ("silent `False` hides shape bugs") is only fixed for the dense `Tensor` path. Comparing e.g. `DistanceMatrix(3)` vs `DistanceMatrix(4)` still returns `False` rather than raising. Matrices can't broadcast, so raising would arguably be the more consistent choice — at minimum worth a deliberate decision, since the PR title advertises a shape-mismatch fix.

- **The compressed-matrix infinity fix (#51) is untested.** `scalar_allclose` now also corrects the `inf` behavior for `DistanceMatrix`/`SymmetricMatrix` (matching-precision compact path), but `test_tensor.cpp` only adds `InfinityComparison` for the dense `Tensor`. Since this is a real behavior change for matrices, a compact-matrix inf case (reflexive `+inf`, `+inf` vs `-inf`) would lock it in. The differing-precision matrix path goes through `np.allclose`, which already handles inf, so that side is covered indirectly.

#### 🟢 Nits

- **`test_allclose.py::test_incompatible_shapes_raise` has a dead variable.** `b = sb.FloatTensor(np.ones((2, 3)))` is constructed and never used; only `a` vs `c` is asserted. The inline comment (`"... -> compatible"` / `"-> incompatible"`) is self-correcting and confusing. Drop `b` or add an assertion that `allclose(a, b)` succeeds (it should: `(3,)` broadcasts against `(2,3)`), which would actually strengthen the test.

- **`test_matches_numpy_isclose` is a weak agreement check.** Both sides evaluate to `False` (driven by the `-inf` vs `inf` and `nan` entries), so it only confirms agreement on a `False` result. A vector that mixes close and not-close entries, or a second assertion on an all-close inf vector, would make the "matches NumPy" claim more meaningful. (Naming nit: it asserts against `np.allclose`, not `np.isclose`.)

- **Misleading variable name** in `test_broadcast_compatible_shapes`: `col = sb.FloatTensor(np.full((1, 3), 2.0))` is shaped `(1, 3)` (a row, not a column). Logic is correct; name is just misleading.

- **Positional `atol`/`rtol` ordering differs from NumPy** (pre-existing, not introduced here): `sb.allclose(a, b, atol, rtol)` vs `np.allclose(a, b, rtol, atol)`. Since the PR explicitly aligns inf/NaN/broadcast semantics with NumPy, a positional caller porting from NumPy would silently swap tolerances. Worth a docstring note, even if changing the order is out of scope.

### Verdict

Solid, well-scoped fix with correct C++ and Python logic and good test coverage for the dense path. No blocking issues — the items above are test-hardening and a consistency call on matrix shape mismatches.


---

PR #158, opened 2026-06-26
Regression tests for already-fixed bug-scan issues (#36, #48, #59, #78, #81)

### LINKED ISSUE #36: [pointprocess] sample_poisson accepts negative rate with no validation, passing negative mean to

### Summary
lambda = rate*volume becomes negative and is passed to std::poisson_distribution<size_t>, whose required precondition is mean > 0; a negative mean is undefined behavior (it merely happens to yield 0 on this libstdc++ build, so the result is non-portable and unvalidated). No validation exists in poisson.py or sample_poisson(). Reported as low because it currently does not crash here, but it is technically UB and a misleading silent no-op.

### Minimal reproduction
```python
from masspcf.point_process import poisson
A = poisson.sample_poisson((5,), dim=2, rate=-3.0, generator=mpcf.random.Generator(1))
print('neg-rate counts:', [np.asarray(A[i]).shape[0] for i in range(5)])  # -> [0, 0, 0, 0, 0], exit 0
```

### Observed
rate=-3.0 (and -0.0001, -1e9, 0.0) returns clouds with 0 points, no error, exit 0. No validation of `rate` exists anywhere in the call chain: masspcf/point_process/poisson.py:sample_poisson passes rate straight to the backend; src/python/point_process/py_poisson.cpp forwards it unchecked; include/mpcf/point_process/poisson.hpp computes lambda = rate*volume (= -3.0 for unit volume) and passes static_cast<double>(lambda) to std::poisson_distribution<size_t>. The C++ standard ([rand.dist.pois.poisson]) requires the distribution's mean parameter to satisfy 0 < mean; a negative mean is undefined behavior. A standalone g++ -O2 -std=c++20 program with std::poisson_distribution<size_t>(-3.0) on this libstdc++ build also returns 0,0,0,0,0 -- confirming the 0-point result is an implementation artifact of UB, not a guaranteed outcome.

### Expected
A negative intensity is meaningless for a Poisson process. The wrapper (consistent with its existing lo<=hi and lo/hi-length validation, which both raise ValueError) should raise a ValueError for rate < 0 rather than constructing a std::poisson_distribution with a negative mean. At minimum it must not rely on undefined behavior; the current silent no-op is non-portable (could loop, assert, or return garbage under a different standard library / hardened build) and contradicts the documented meaning of `rate` as the Poisson intensity lambda.

### Root-cause hint
masspcf/point_process/poisson.py:sample_poisson (and include/mpcf/point_process/poisson.hpp:mpcf::pp::sample_poisson) accept `rate` with no `rate >= 0` check, so a negative lambda reaches std::poisson_distribution whose precondition is 0 < mean (UB).

In `test/python/test_bugscan_pointprocess.py` -- run with `cd test && python -m pytest python/test_bugscan_pointprocess.py`

```python
# Bug 35: sample_poisson accepts a negative rate with no validation, passing a
# negative mean to std::poisson_distribution (UB / silent no-op).
# Non-crashing here -> normal in-process test.
def test_sample_poisson_negative_rate_raises():
    """A negative Poisson intensity must be rejected with a clean ValueError."""
    # BUG: sample_poisson accepts negative rate with no validation, passing a
    #      negative mean to std::poisson_distribution (undefined behavior).
    # Expected: ValueError for rate < 0, consistent with the existing lo<=hi /
    #           lo-hi-length validation (test_lo_greater_than_hi_raises).
    # Observed today: returns clouds with 0 points, no error, exit 0 -- relying
    #           on UB that merely happens to yield empty clouds on this build.
    mpcf.system.force_cpu(True)
    try:
            sample_poisson((5,), dim=2, rate=-3.0,
                           generator=mpcf.random.Generator(1))
    finally:
        mpcf.system.force_cpu(False)
```

</details>

### LINKED ISSUE #48: [interop] Ordering comparisons (<, <=, >, >=) against a scalar or ndarray leak AttributeError about internal

### Summary
Wrong-kind/leaky internal error on ordinary input; inconsistent with == which handles ndarray. Ordering against scalars/ndarrays is a core NumPy-like operation.

### Minimal reproduction
```python
a = mpcf.FloatTensor(np.array([1.0, 2.0, 3.0]))
try:
    a < 2.0
    print("no error")
except AttributeError as e:
    print("AttributeError:", e)
# -> prints: AttributeError: 'float' object has no attribute '_data'
# Same with: a < np.array([2.0, 2.0, 2.0]) -> 'numpy.ndarray' object has no attribute '_data'
# Also reflected: 2.0 < a -> same AttributeError
```

### Observed
__lt__/__le__/__gt__/__ge__ in masspcf/_tensor_base.py do `rhs._data` unconditionally with no type guard. `FloatTensor([1,2,3]) < 2.0` raises `AttributeError: 'float' object has no attribute '_data'`; `FloatTensor([1,2,3]) < np.array([2,2,2])` raises `AttributeError: 'numpy.ndarray' object has no attribute '_data'`. Reflected forms (e.g. `2.0 < a`) raise the same. Tensor-vs-tensor comparison (`a < b` where both are FloatTensor) works fine and returns the correct BoolTensor. Inconsistency confirmed: `a == np.array([1,2,3])` succeeds and returns BoolTensor([True,True,True]) because __eq__ returns NotImplemented for non-Tensor (deferring to numpy via __array__), whereas the ordering ops crash instead of returning NotImplemented.

### Expected
Either an elementwise BoolTensor (NumPy semantics: a < 2.0 -> [True, False, False]) or, at minimum, a clean TypeError / NotImplemented -- not a leaked internal AttributeError about a private '_data' attribute. __eq__/__ne__ already guard with `if not isinstance(rhs, Tensor): return NotImplemented`, and arithmetic ops use the `_decay_operand` helper (`val._data if hasattr(val, "_data") else val`); the ordering ops should follow one of those patterns rather than blindly accessing rhs._data.

### Root-cause hint
masspcf/_tensor_base.py:__lt__/__le__/__gt__/__ge__ (lines 592-606): each does `self._data <OP> rhs._data` with no type check, so any non-Tensor rhs (Python scalar, numpy ndarray) leaks AttributeError; unlike __eq__/__ne__ which return NotImplemented for non-Tensor, and unlike arithmetic ops which route rhs through _decay_operand.

```python
# Bug 47 + Bug 48: ordered comparisons (<, <=, >, >=) against a scalar leak a
#         raw AttributeError about the internal '_data' attribute.
def test_ordered_comparison_scalar_returns_bool_tensor():
    """< <= > >= against a scalar must broadcast to a BoolTensor, not leak _data."""
    # BUG: ordered ops against a scalar raise AttributeError: 'float' object has
    #      no attribute '_data' (unconditional rhs._data access, no promotion)
    # Expected: elementwise BoolTensor via scalar broadcasting, per docs/arithmetic.rst.
    # Observed today: AttributeError leaking the private '_data' attribute name.
    a = mpcf.FloatTensor(np.array([1.0, 2.0, 3.0]))
    ref = np.array([1.0, 2.0, 3.0])

    for op, npop in [
        (lambda x: x < 2.0, lambda x: x < 2.0),
        (lambda x: x <= 2.0, lambda x: x <= 2.0),
        (lambda x: x > 2.0, lambda x: x > 2.0),
        (lambda x: x >= 2.0, lambda x: x >= 2.0),
    ]:
        r = op(a)
        assert isinstance(r, mpcf.BoolTensor), (
            f"expected BoolTensor, got {type(r).__name__}={r!r}")
        np.testing.assert_array_equal(np.asarray(r), npop(ref))
```

</details>

### LINKED ISSUE #59: [friction/indexing] Writing into a broadcast view silently scatters to multiple elements (no read-only guard like

### Scenario
User broadcasts a column to a wider shape with broadcast_to and then assigns to a single element, expecting either an error (as NumPy gives) or per-element semantics.

### Current approach
```
tab = FloatTensor(np.arange(4.0).reshape(4,1)).broadcast_to((4,3))
tab[0,0] = 99.0   # succeeds; row 0 becomes [99, 99, 99] because all 3 columns alias one stride-0 cell
```

### Desired API
Either raise (NumPy: 'ValueError: assignment destination is read-only' for broadcast arrays) or document the aliasing clearly at the __setitem__/broadcast_to boundary. Silently turning a single-element store into a multi-element scatter is surprising and a likely source of subtle data bugs.

### Notes
broadcast_to docstring does say it shares storage; the friction is that __setitem__ does not warn/guard, unlike NumPy.

### LINKED ISSUE #78: [friction/kernel] np.asarray(SymmetricMatrix) silently produces a 0-d object array instead of the dense matrix

### Scenario
A user gets a kernel from mpcf.l2_kernel and wants to feed it to numpy/scipy (e.g. eigvalsh, a heatmap, sklearn). The natural call is np.asarray(K) or np.array(K).

### Current approach
```
K = mpcf.l2_kernel(X)
a = np.asarray(K)
# a -> array(SymmetricMatrix_f64(size=2), dtype=object), shape (), a.item() is K
# silently wrong; the correct path is the non-obvious K.to_dense()
```

### Desired API
SymmetricMatrix should implement __array__ (and ideally the buffer/array-interface) so np.asarray(K) returns the dense (n, n) float array, matching K.to_dense(). At minimum it should raise rather than return a 0-d object wrapper.

### Notes
Confirmed: np.asarray(K) returns dtype=object shape () wrapping the SymmetricMatrix. This is a silent-data-shape footgun for anyone integrating with the numpy/scipy ecosystem.

### LINKED ISSUE #81: [friction/reductions] Reductions do not accept negative dim (no NumPy-style axis-from-end), inconsistent with the

### Scenario
A user with an (m, n) PcfTensor wants 'reduce over the last axis' without knowing the rank, the natural NumPy idiom mpcf.mean(A, dim=-1). They also expect it because element indexing (A[-1]) and stack(axis=-1) already accept negatives.

### Current approach
```
mpcf.mean(A, dim=-1) raises a confusing pybind TypeError. The user must compute the positive axis manually: ndim = len(A.shape); r = mpcf.mean(A, dim=ndim-1). squeeze(axis=-1) is similarly broken, while stack(axis=-1) and indexing accept negatives, so the behavior is inconsistent across the package.
```

### Desired API
mean(fs, dim=-1) and max_time(fs, dim=-1) should normalize dim via dim % ndim (NumPy semantics), so dim=-1 means the last axis, dim=-2 the second-to-last, etc., and raise numpy.AxisError for genuinely out-of-range values.

### Notes
Fixing the dim-validation bug (changing size_t to a signed type plus a normalize+bounds-check in the Python wrapper or binding) would resolve both the crashes and this friction in one stroke.

bwehlin:
## Summary

These five bug-scan issues were already fixed in earlier work but lacked an issue-tagged regression test. This PR adds one — **tests only, no source changes** — grouped in `test/python/test_bugscan_resolved.py` for traceability:

- **#36** `sample_poisson` rejects a negative `rate` with `ValueError` (and `rate=0` stays valid).
- **#48** ordering comparisons (`<`, `<=`, `>`, `>=`) against a scalar or ndarray return a `BoolTensor` (no `AttributeError` about internal `_data`), including the reflected form.
- **#59** writing through a broadcast view (element or in-place) raises `ValueError`; the view stays readable and `.copy()` is writeable.
- **#78** `np.asarray` of an `l2_kernel` result (a `SymmetricMatrix`) is a dense `(n, n)` array, not a 0-d object array.
- **#81** `mean` / `max_time` accept a negative `dim` (matching the positive index) and raise on out-of-range.

All 16 tests pass; the full Python suite is unaffected.

Closes #36
Closes #48
Closes #59
Closes #78
Closes #81

### DIFF
--- a/test/python/test_bugscan_resolved.py
+++ b/test/python/test_bugscan_resolved.py
@@ -1,0 +1,119 @@
+"""Regression tests for bug-scan issues that were already fixed before this
+sweep but lacked an issue-tagged test. Grouped here for traceability.
+
+Covers #36, #48, #59, #78, #81.
+"""
+import numpy as np
+import pytest
+
+import stablebear as sb
+import stablebear.point_process as pp
+from stablebear.reductions import mean, max_time
+
+
+# ---------------------------------------------------------------------------
+# #36: sample_poisson with a negative rate passed a negative mean to
+# std::poisson_distribution (UB). It now validates rate and raises ValueError.
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("rate", [-3.0, -0.5, -1e9])
+def test_sample_poisson_negative_rate_raises(rate):
+    with pytest.raises(ValueError):
+        pp.sample_poisson((3,), dim=2, rate=rate, generator=sb.random.Generator(1))
+
+
+def test_sample_poisson_zero_rate_is_valid():
+    # rate == 0 is a valid (degenerate) input: empty clouds, no error.
+    out = pp.sample_poisson((2,), dim=2, rate=0.0, generator=sb.random.Generator(1))
+    assert out.shape == (2,)
+
+
+# ---------------------------------------------------------------------------
+# #48: ordering comparisons (<, <=, >, >=) against a scalar or ndarray used to
+# leak an AttributeError about the internal _data. They now return a BoolTensor.
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("op, ref", [
+    (lambda a: a < 2.0, [True, False, False]),
+    (lambda a: a <= 2.0, [True, True, False]),
+    (lambda a: a > 2.0, [False, False, True]),
+    (lambda a: a >= 2.0, [False, True, True]),
+])
+def test_ordering_against_scalar_returns_bool_tensor(op, ref):
+    a = sb.FloatTensor([1.0, 2.0, 3.0])
+    result = op(a)
+    assert isinstance(result, sb.BoolTensor)
+    np.testing.assert_array_equal(np.asarray(result), np.array(ref))
+
+
+def test_ordering_reflected_and_ndarray():
+    a = sb.FloatTensor([1.0, 2.0, 3.0])
+    assert isinstance(2.0 < a, sb.BoolTensor)
+    np.testing.assert_array_equal(np.asarray(2.0 < a), np.array([False, False, True]))
+    res = a < np.array([2.0, 2.0, 2.0])
+    np.testing.assert_array_equal(np.asarray(res), np.array([True, False, False]))
+
+
+# ---------------------------------------------------------------------------
+# #59: writing through a broadcast view silently scattered to multiple elements
+# (no read-only guard). Writing through a broadcast view now raises ValueError.
+# ---------------------------------------------------------------------------
+
+
+def test_column_broadcast_view_is_read_only():
+    src = sb.FloatTensor(np.arange(4.0).reshape(4, 1))
+    view = src.broadcast_to((4, 3))
+    with pytest.raises(ValueError):
+        view[0, 0] = 99.0
+    with pytest.raises(ValueError):
+        view += 1.0
+    # the view is still readable, and .copy() yields a writeable tensor
+    np.testing.assert_array_equal(np.asarray(view)[:, 0], np.arange(4.0))
+    writeable = view.copy()
+    writeable[0, 0] = 99.0
+    assert np.asarray(writeable)[0, 0] == 99.0
+
+
+# ---------------------------------------------------------------------------
+# #78: np.asarray(SymmetricMatrix) produced a 0-d object array. It now returns
+# the dense (n, n) matrix -- in particular for an l2_kernel result.
+# ---------------------------------------------------------------------------
+
+
+def test_l2_kernel_asarray_is_dense():
+    X = sb.PcfTensor([
+        sb.Pcf(np.array([[0.0, 1.0], [1.0, 2.0]])),
+        sb.Pcf(np.array([[0.0, 2.0], [1.0, 3.0]])),
+        sb.Pcf(np.array([[0.0, 0.0], [1.0, 1.0]])),
+    ])
+    K = sb.l2_kernel(X)
+    arr = np.asarray(K)
+    assert arr.dtype != object
+    assert arr.shape == (3, 3)
+    # symmetric kernel
+    np.testing.assert_allclose(arr, arr.T)
+
+
+# ---------------------------------------------------------------------------
+# #81: reductions (mean / max_time) accept a negative dim (NumPy axis-from-end).
+# ---------------------------------------------------------------------------
+
+
+def test_mean_negative_dim_matches_positive():
+    t = sb.zeros((2, 3, 4))
+    assert tuple(mean(t, dim=-1).shape) == tuple(mean(t, dim=2).shape) == (2, 3)
+    assert tuple(mean(t, dim=-2).shape) == tuple(mean(t, dim=1).shape) == (2, 4)
+
+
+def test_max_time_negative_dim_matches_positive():
+    t = sb.zeros((2, 3, 4))
+    assert tuple(max_time(t, dim=-1).shape) == tuple(max_time(t, dim=2).shape)
+
+
+@pytest.mark.parametrize("bad_dim", [3, -4, 100])
+def test_reduction_out_of_range_dim_raises(bad_dim):
+    t = sb.zeros((2, 3, 4))
+    with pytest.raises(IndexError):   # numpy.AxisError subclasses IndexError
+        mean(t, dim=bad_dim)


bwehlin:
## 🔍 Adversarial review

### Summary
This is a tests-only PR adding `test/python/test_bugscan_resolved.py` (16 tests) as issue-tagged regression coverage for five already-fixed bug-scan issues (#36, #48, #59, #78, #81). I cross-checked every test against the actual source and the claimed behavior holds in each case. **Verdict: solid.** I tried to break it and found no correctness problems — only a handful of minor nits.

I confirmed against the source that each test actually exercises the fix it claims:
- **#36** — `rate` validation lives in C++ (`include/sbear/point_process/poisson.hpp:34`, `throw std::invalid_argument` → pybind11 maps to `ValueError`); the Python wrapper does no validation, so the test correctly drives the real guard. `rate=0` → `lambda=0` → `std::poisson_distribution(0)` always yields 0, so the "valid degenerate" assertion is right.
- **#48** — `_compare` (`stablebear/_tensor_base.py:623`) returns `BoolTensor` for scalar/ndarray RHS; reflected `2.0 < a` dispatches to `a.__gt__` since `FloatTensor` isn't a `float` subclass, so the `[False, False, True]` expectation is correct.
- **#59** — the guard is `_ensure_writeable` (`stablebear/_tensor_base.py:455-471`), called from both `__setitem__` and the in-place ops; the `(4,1)→(4,3)` view has a size-3 stride-0 axis so both `view[0,0]=` and `view += 1.0` raise `ValueError`. `.copy()` clears stride-0, staying writeable. Correct.
- **#78** — `SymmetricMatrix.__array__` (`stablebear/symmetric_matrix.py:148`) returns `to_dense()`; `l2_kernel` returns a `SymmetricMatrix` (`stablebear/inner_product.py:13`), so `np.asarray(K)` is dense `(n,n)`. The three `Pcf` inputs are valid `(n,2)` arrays starting at `t=0`.
- **#81** — `_resolve_dim` (`stablebear/reductions.py:26-39`) normalizes negative dims and raises `IndexError` out of range; `sb.zeros` defaults to `pcf32`, so the inputs are valid for `mean`/`max_time`.

### 🟢 Nits
- **`test_bugscan_resolved.py:117`** — the comment `# numpy.AxisError subclasses IndexError` is misleading. `_resolve_dim` raises a **plain** `IndexError` (`reductions.py:36`), not `np.AxisError`. The `pytest.raises(IndexError)` assertion passes either way, but the comment implies a numpy code path that isn't taken.
- **`test_column_broadcast_view_is_read_only` (#59, line ~70)** substantially overlaps the existing `test/python/test_bugscan_broadcast.py`, which already covers element writes, in-place ops (incl. tensor RHS), prepended-axis broadcasts, IntTensor, the readable/`.copy()` invariants, and the size-1 stride-0 exception. The only genuinely new coverage here is the *trailing-axis* `(4,1)→(4,3)` broadcast; the rest is redundant. Not a problem, just worth knowing for traceability.
- **`test_ordering_reflected_and_ndarray` (#48)** only checks the reflected form for `<` (`2.0 < a`). Reflected `<=`/`>`/`>=` (e.g. `2.0 >= a`) go through the same `_compare`/operator-swap path so they're very likely fine, but they're untested.
- Minor: none of these tests are marked GPU/CPU-specific. They should be backend-agnostic (all exercise small CPU paths), so this is just an observation, not a concern.


---

test/python/test_bugscan_resolved.py
```diff
@@ -0,0 +1,119 @@
...
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("rate", [-3.0, -0.5, -1e9])
+def test_sample_poisson_negative_rate_raises(rate):
+    with pytest.raises(ValueError):
+        pp.sample_poisson((3,), dim=2, rate=rate, generator=sb.random.Generator(1))
+
+
+def test_sample_poisson_zero_rate_is_valid():
```
MattiasBjerklov, line 26: Compare with the comment https://github.com/kthtda/stablebear/issues/36#issuecomment-4916337123

test/python/test_bugscan_resolved.py
```diff
@@ -0,0 +1,119 @@
...
+    assert np.asarray(writeable)[0, 0] == 99.0
+
+
+# ---------------------------------------------------------------------------
+# #78: np.asarray(SymmetricMatrix) produced a 0-d object array. It now returns
+# the dense (n, n) matrix -- in particular for an l2_kernel result.
+# ---------------------------------------------------------------------------
+
+
+def test_l2_kernel_asarray_is_dense():
```
MattiasBjerklov, line 85: The l2_kernel contains inf. The test runs anyway, but perhaps this is not quite clean.

MattiasBjerklov:
Tests look ok, but see the comments

PR #202, opened 2026-07-02
Fix CUDA pipeline robustness: data race, swallowed exceptions, null-deref, uninterruptible runs

### LINKED ISSUE #187: [cuda] Null-pointer deref (segfault) when the CUDA backend loads but zero usable devices exist at runtime
## Summary

The CUDA dispatch branch in the bindings guards only on the factory, force-CPU flag, and threshold:

```cpp
// src/python/functional/py_distance.cpp:55 (also :172), py_inner_product.cpp:48
if (cudaFactory && !sb::settings().forceCpu
    && static_cast<size_t>(std::distance(begin, end)) >= sb::settings().cudaThreshold)
```

It never checks that `default_executor().cuda()` is non-null before the factory path dereferences it (`src/cuda/cuda_matrix_integrate.cu:21` and `:78`: `*default_executor().cuda()`).

`Executor` (`include/sbear/executor.hpp:31-41`) only creates the CUDA executor when `get_num_cuda_devices() > 0`, and `src/executor.cpp:21-23` returns 0 on **any** `cudaGetDeviceCount` failure.

## Failure scenario

GPU detection succeeds (sysfs/lspci sees an NVIDIA device), so `_sb_cudaXX` is selected and imports fine — importing succeeds even without a working driver because cudart initializes lazily. At runtime there are zero usable devices (missing/mismatched driver, `CUDA_VISIBLE_DEVICES=""`, exhausted MIG contexts...). The user calls `pdist`/`cdist`/`l2_kernel` with ≥ `cudaThreshold` (default 500) PCFs → dereference of a null `tf::Executor*` → **segfault**, instead of the obvious correct behavior: fall back to the CPU path.

## Suggested fix

Add `&& sb::default_executor().cuda() != nullptr` to the dispatch condition (all three sites), so such systems silently take the CPU path (optionally with a `deviceVerbose` note).

### LINKED ISSUE #188: [cuda] Exceptions inside the CUDA block pipeline are silently swallowed — Python receives a partially-zero distance matrix with no error
## Summary

`include/sbear/cuda/cuda_block_executor.cuh:143-144`:

```cpp
m_gpuThreads.run(std::move(flow));
m_gpuThreads.wait_for_all();
```

The `tf::Future` returned by `run()` is discarded, and `wait_for_all()` does **not** rethrow task exceptions. Every `CHK_CUDA` throw inside `exec_block` — kernel launch failure, device OOM in `upload_subset`, invalid launch configuration (e.g. `gridDim.y > 65535` for very wide blocks in `get_grid_dims`, `block_matrix_support.cuh:26-35`) — is captured in the discarded future; Taskflow then just cancels the remaining tasks of that topology.

`pipeline.execute()` returns normally, the outer future completes cleanly, and `StoppableTask::wait_for` (which correctly rethrows from the *outer* future, `task.hpp:84-96`) reports success.

## Failure scenario

Any CUDA runtime error mid-computation (most realistically: device OOM on a large `pdist`) → Python receives a distance matrix in which the blocks after the failure are **silently zero**, with no exception, no warning, nothing. Downstream statistics are quietly wrong. The CPU path does not have this problem — its exceptions flow into the retained outer future.

## Suggested fix

Retain the future and rethrow: `m_gpuThreads.run(std::move(flow)).get();` (Taskflow futures propagate the first task exception), or collect per-task `std::exception_ptr`s in `exec_block` and rethrow after the wait. Either way, a CUDA failure must surface to the caller.

### LINKED ISSUE #189: [cuda] Data race in CudaOffsetDataManager::upload_subset on multi-GPU systems — shared mutable m_localOffsets written by all GPU workers
## Summary

`include/sbear/cuda/cuda_offset_data_manager.cuh:36-43` — `upload_subset` is `const` but resizes and writes the shared `mutable std::vector<size_t> m_localOffsets` scratch buffer with no synchronization:

```cpp
m_localOffsets.resize(count + 1);
for (size_t i = 0; i <= count; ++i)
  m_localOffsets[i] = hostData.offsets[start + i] - baseOffset;
deviceOffsets.toDevice(m_localOffsets.data(), count + 1);
```

The same manager instance is shared by **all** GPU worker threads:
- `CudaBlockPipeline::execute` runs `exec_block` via `for_each_index` on the CUDA executor, one worker per GPU (`cuda_block_executor.cuh:135-143`);
- each `exec_block` calls `m_rowDataManager.upload_subset(...)` / `m_colDataManager.upload_subset(...)` on the single shared `CudaPcfDataManager` (`pcf_block_op.cuh:90-96`, members at 117-118; the pdist task passes the *same* manager for rows and cols).

## Failure scenario

On a machine with ≥ 2 GPUs, two workers concurrently resize/write the same vector → torn or wrong offset tables uploaded to the device → **silently wrong distance matrices**, or OOB reads in the kernel when a corrupted offset exceeds the element buffer. Single-GPU machines are unaffected, which is likely why this hasn't surfaced (CI's self-hosted runner tests 1 GPU).

Found independently by two separate review passes over this code.

## Suggested fix

Make `m_localOffsets` a local variable of `upload_subset` (it's pure scratch — the allocation-reuse micro-optimization isn't worth the race), or make it `thread_local`, or give each GPU worker its own manager instance.

### LINKED ISSUE #195: [cuda] request_stop() / Ctrl-C cannot interrupt a running CUDA computation — pipeline.cancel() is never wired up
## Summary

`CudaBlockPipeline` has a `cancel()` method (sets `m_canceled`, which `exec_block` checks per block), but **nothing ever calls it** — grep finds zero call sites. The only stop check in the CUDA tasks happens *before* `pipeline.execute()` (`pcf_block_op.cuh:186, 268`), so once execution starts, a stop request is ignored until the whole matrix is done.

## Failure scenario

User hits Ctrl-C during a large GPU `pdist`. `stablebear/async_task.py:41-48` responds by calling `task.request_stop()` and then draining — which, for the GPU path, blocks until the **entire** distance matrix finishes (potentially minutes). The same interrupt on the CPU path aborts within one row. To the user it looks like Ctrl-C is broken, and the escape hatch is kill -9-ing the interpreter.

## Suggested fix

Override `on_stop_requested()` in the CUDA tasks to call `pipeline.cancel()` (the per-block cancellation check in `exec_block` already exists and would make interruption take effect at block granularity).

bwehlin:
Four CUDA runtime-robustness fixes from the bug scan, grouped because they touch the same pipeline files (`cuda_block_executor.cuh`, `pcf_block_op.cuh`) and dispatch sites. One commit per issue.

- **#189 (high) — multi-GPU data race**: `CudaOffsetDataManager::upload_subset` resized and wrote a shared `mutable` scratch vector from all GPU worker threads with no synchronization, producing torn offset tables (silently wrong distance matrices or OOB kernel reads) on 2+ GPU machines. The scratch buffer is now a local variable.
- **#188 (high) — CUDA errors silently swallowed**: `CudaBlockPipeline::execute` discarded the `tf::Future` from `run()` and used `wait_for_all()`, which does not rethrow task exceptions — a mid-computation CUDA failure (e.g. device OOM) returned a partially-zero matrix with no error. Now the future is retained and `.get()` rethrows into the outer task future that `StoppableTask::wait_for` already propagates to Python.
- **#187 (high) — segfault with zero usable CUDA devices**: all three CUDA dispatch sites (`pdist`/`cdist`/`l2_kernel`) now check `default_executor().cuda() != nullptr` and fall back to the CPU path instead of dereferencing null when the CUDA module loads but no device is usable at runtime (driver mismatch, `CUDA_VISIBLE_DEVICES=""`, ...).
- **#195 (medium) — Ctrl-C couldn't interrupt CUDA**: `CudaBlockPipeline::cancel()` existed but had no call sites, so `request_stop()` during a GPU `pdist` blocked until the whole matrix finished. The pipeline now polls a stop predicate per block; the CUDA tasks pass `StoppableTask::stop_requested()`, so interruption takes effect at block granularity like the CPU path.

Fixes #189, fixes #188, fixes #187, fixes #195

## Tests

New `test/test_cuda_block_pipeline.cu` (runs in the `BUILD_CUDA_TESTER` suite on the GPU runner), driving the real `CudaBlockPipeline` with stub BlockOps:

- an exception thrown inside `exec_block` propagates out of `execute()` (#188 — fails on the pre-fix `wait_for_all` code);
- a stop request skips all blocks (#195), and `cancel()` still does too;
- the normal path executes every scheduled block.

The #189 race itself has no deterministic single-run test (needs ≥2 GPUs plus a sanitizer); #187 needs an environment with a broken driver to trigger.

## Verification

No GPU or CUDA toolkit is available in the sandbox, so these changes were not executed on a device here: verified by syntax checks and review of the exception-propagation chain (per the contract documented in `task.hpp:78-96`, which the CPU path already relies on). Full CPU C++ + Python suites pass with these changes applied. The new `.cu` tests will exercise #188/#195 on the self-hosted GPU runner; a multi-GPU run for #189 before release is still recommended.

### DIFF
--- a/include/sbear/cuda/cuda_block_executor.cuh
+++ b/include/sbear/cuda/cuda_block_executor.cuh
@@ -140,8 +140,10 @@ namespace sb
         exec_block(blocks[i], blockDim);
       });
 
-      m_gpuThreads.run(std::move(flow));
-      m_gpuThreads.wait_for_all();
+      // wait_for_all() discards task exceptions; get() rethrows the first
+      // CHK_CUDA failure (OOM, bad launch, ...) so it reaches the caller
+      // instead of silently leaving the remaining blocks zeroed.
+      m_gpuThreads.run(std::move(flow)).get();
 
       // Flush any pending scatter on each GPU
       for (size_t iGpu = 0; iGpu < m_nGpus; ++iGpu)
--- a/include/sbear/cuda/cuda_block_executor.cuh
+++ b/include/sbear/cuda/cuda_block_executor.cuh
@@ -114,12 +114,14 @@ namespace sb
         BlockOp& blockOp,
         const CudaBlockScheduler& scheduler,
         ResultWriter& writer,
-        std::function<void(size_t)> progressCb = [](size_t) {})
+        std::function<void(size_t)> progressCb = [](size_t) {},
+        std::function<bool()> stopRequested = [] { return false; })
       : m_gpuThreads(gpuThreads)
       , m_blockOp(blockOp)
       , m_scheduler(scheduler)
       , m_writer(writer)
       , m_progressCb(progressCb)
+      , m_stopRequested(stopRequested)
       , m_nGpus(gpuThreads.num_workers())
     {
       m_canceled.store(false);
@@ -133,7 +135,10 @@ namespace sb
 
       tf::Taskflow flow;
       flow.for_each_index<size_t, size_t, size_t>(0ul, blocks.size(), 1ul, [this, &blocks, blockDim](size_t i) {
-        if (m_canceled.load())
+        // Checked per block so a stop request (e.g. Ctrl-C via
+        // StoppableTask::request_stop) interrupts a running computation at
+        // block granularity instead of after the whole matrix.
+        if (m_canceled.load() || m_stopRequested())
         {
           return;
         }
@@ -336,6 +341,7 @@ namespace sb
     const CudaBlockScheduler& m_scheduler;
     ResultWriter& m_writer;
     std::function<void(size_t)> m_progressCb;
+    std::function<bool()> m_stopRequested;
     size_t m_nGpus;
 
     std::vector<GpuBlockStorage> m_gpuStorages;
--- a/src/python/functional/py_distance.cpp
+++ b/src/python/functional/py_distance.cpp
@@ -52,7 +52,12 @@ namespace
       auto end = sb::end1dValues(fs);
 
 #ifdef BUILD_WITH_CUDA
+      // default_executor().cuda() is null when the CUDA module loaded but no
+      // usable device exists at runtime (e.g. driver mismatch or
+      // CUDA_VISIBLE_DEVICES=""); fall back to the CPU path instead of
+      // dereferencing it in the factory.
       if (cudaFactory && !sb::settings().forceCpu
+          && sb::default_executor().cuda() != nullptr
           && static_cast<size_t>(std::distance(begin, end)) >= sb::settings().cudaThreshold)
       {
         if (sb::settings().deviceVerbose)
@@ -170,6 +175,7 @@ namespace
 
 #ifdef BUILD_WITH_CUDA
       if (cudaFactory && !sb::settings().forceCpu
+          && sb::default_executor().cuda() != nullptr
           && xTotal * yTotal >= sb::settings().cudaThreshold * sb::settings().cudaThreshold)
       {
         if (sb::settings().deviceVerbose)
--- a/test/test_cuda_block_pipeline.cu
+++ b/test/test_cuda_block_pipeline.cu
@@ -1,0 +1,148 @@
+// Regression tests for the CUDA block pipeline (issues #188 and #195):
+// - exceptions thrown inside exec_block must propagate out of execute()
+//   instead of being swallowed by wait_for_all (issue #188);
+// - a stop request must interrupt the pipeline at block granularity
+//   (issue #195), and cancel() must keep working.
+//
+// Uses stub BlockOps against the real pipeline so the behavior is exercised
+// without depending on kernel contents. Requires a CUDA device (the pipeline
+// allocates device buffers up front).
+
+#include <gtest/gtest.h>
+
+#ifdef BUILD_WITH_CUDA
+
+#include <sbear/cuda/cuda_block_executor.cuh>
+#include <sbear/cuda/cuda_block_scheduler.hpp>
+#include <sbear/executor.hpp>
+
+#include <atomic>
+#include <stdexcept>
+
+namespace
+{
+
+  struct NullWriter
+  {
+    void scatter(const float*, const sb::BlockInfo&) {}
+  };
+
+  // exec_block launches nothing; it just counts invocations.
+  struct CountingBlockOp
+  {
+    struct GpuStorage {};
+
+    GpuStorage init_gpu_storage(size_t, const sb::CudaBlockScheduler&) { return {}; }
+
+    void exec_block(GpuStorage&, const sb::BlockInfo&, sb::CudaDeviceArray<float>&, dim3)
+    {
+      ++calls;
+    }
+
+    std::atomic<int> calls{0};
+  };
+
+  struct ThrowingBlockOp
+  {
+    struct GpuStorage {};
+
+    GpuStorage init_gpu_storage(size_t, const sb::CudaBlockScheduler&) { return {}; }
+
+    void exec_block(GpuStorage&, const sb::BlockInfo&, sb::CudaDeviceArray<float>&, dim3)
+    {
+      throw std::runtime_error("simulated CUDA failure");
+    }
+  };
+
+  class CudaBlockPipelineTest : public ::testing::Test
+  {
+  protected:
+    void SetUp() override
+    {
+      if (sb::get_num_cuda_devices() == 0)
+      {
+        GTEST_SKIP() << "No CUDA devices available";
+      }
+    }
+
+    static sb::CudaBlockScheduler make_scheduler()
+    {
+      return sb::CudaBlockScheduler({
+          .nRows = 8, .nCols = 8,
+          .maxOutputElements = 16,
+          .nSplitsHint = 4,
+          .triangleMode = sb::BlockTriangleMode::Full,
+          .minBlockSide = 2
+      });
+    }
+  };
+
+  TEST_F(CudaBlockPipelineTest, ExecutesEveryBlock)
+  {
+    auto scheduler = make_scheduler();
+    ASSERT_GT(scheduler.blocks().size(), 1u);
+
+    tf::Executor gpuThreads(1);
+    CountingBlockOp op;
+    NullWriter writer;
+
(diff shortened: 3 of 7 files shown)


include/sbear/cuda/cuda_block_executor.cuh
```diff
@@ -133,15 +135,20 @@ namespace sb
...
         }
         exec_block(blocks[i], blockDim);
       });
 
-      m_gpuThreads.run(std::move(flow));
-      m_gpuThreads.wait_for_all();
+      // wait_for_all() discards task exceptions; get() rethrows the first
+      // CHK_CUDA failure (OOM, bad launch, ...) so it reaches the caller
+      // instead of silently leaving the remaining blocks zeroed.
+      m_gpuThreads.run(std::move(flow)).get();
```
bwehlin, line 151: **[medium] New exception path can free pinned host memory with an async D2H copy still in flight.**

When `.get()` rethrows, `execute()` now exits before the `flush_pending_scatter` loop below, so `downloadStream` is never synchronized. Realistic failure sequence: block N completes and enqueues its async D2H via `toHostAsync` (sets `hasPendingScatter`), block N+1 then throws (e.g. device OOM in `upload_subset`). `.get()` waits for the topology and rethrows; the stack unwinds and the pipeline is destroyed while the copy of block N may still be running:

- `~GpuBlockStorage` calls `cudaStreamDestroy`, which is documented to return immediately while the device keeps executing the stream's work;
- then the `hostScratch` `PinnedHostBuffer` destructors call `cudaFreeHost` on buffers the in-flight copy is writing into.

The only thing preventing a write-after-free today is member destruction order: `opStorage` is destroyed first and its `cudaFree` calls *usually* implicitly synchronize the device — undocumented behavior you shouldn't rely on (it's exactly why `cudaFreeAsync` exists).

Pre-fix this path couldn't occur (the exception was swallowed and the flush loop always ran), so this is introduced by the PR. Suggest:

```cpp
try {
  m_gpuThreads.run(std::move(flow)).get();
} catch (...) {
  for (size_t iGpu = 0; iGpu < m_nGpus; ++iGpu) {
    if (m_gpuStorages[iGpu].downloadStream)
      (void)cudaStreamSynchronize(m_gpuStorages[iGpu].downloadStream);
  }
  throw;
}
```

Note the new GTest can't catch this: `ThrowingBlockOp` throws on the very first block, so there is never a pending D2H at unwind time.

---

test/test_cuda_block_pipeline.cu
```diff
@@ -0,0 +1,148 @@
...
+    tf::Executor gpuThreads(1);
+    CountingBlockOp op;
+    NullWriter writer;
+
+    sb::CudaBlockPipeline<float, CountingBlockOp, NullWriter> pipeline(
+        gpuThreads, op, scheduler, writer,
+        [](size_t) {}, [] { return true; });
+    pipeline.execute(dim3(8, 8, 1));
+
+    EXPECT_EQ(op.calls.load(), 0);
```
bwehlin, line 127: **[low] This test doesn't pin the per-block polling that #195 claims.**

With a constantly-`true` predicate, the test only proves the predicate is consulted at least once before any block runs. An implementation that checked stop a single time up front and then ran the whole matrix uninterrupted (i.e. the pre-PR behavior, where the only check was `if (stop_requested()) return;` before `pipeline.execute()` in `pcf_block_op.cuh`) would pass this test equally well — `calls` would still be 0.

To actually lock in block-granularity interruption, flip the predicate after the first executed block, e.g.:

```cpp
sb::CudaBlockPipeline<float, CountingBlockOp, NullWriter> pipeline(
    gpuThreads, op, scheduler, writer,
    [](size_t) {}, [&op] { return op.calls.load() >= 1; });
pipeline.execute(dim3(8, 8, 1));

EXPECT_EQ(op.calls.load(), 1);  // interrupted mid-run, not after the whole matrix
```

(Single-worker executor makes this deterministic.)

---

src/python/functional/py_distance.cpp
```diff
@@ -52,7 +52,12 @@ namespace
       auto end = sb::end1dValues(fs);
 
 #ifdef BUILD_WITH_CUDA
+      // default_executor().cuda() is null when the CUDA module loaded but no
+      // usable device exists at runtime (e.g. driver mismatch or
+      // CUDA_VISIBLE_DEVICES=""); fall back to the CPU path instead of
+      // dereferencing it in the factory.
       if (cudaFactory && !sb::settings().forceCpu
+          && sb::default_executor().cuda() != nullptr
```
bwehlin, line 60: **[low] User-visible behavior changes are undocumented (CLAUDE.md requires docs in the same commit).**

Two behaviors in this PR are user-visible and not reflected in `docs/`:

1. `docs/gpu.rst` ("Supported platforms") currently documents CPU fallback only for the case where "the CUDA runtime library cannot be loaded" (with a warning). This PR adds a second, fully *silent* fallback: module loads fine but zero usable devices at runtime (driver mismatch, `CUDA_VISIBLE_DEVICES=""`, ...). A user with `set_device_verbose(True)` just sees "on CPU(s)" with no hint why the GPU was skipped — issue #187 explicitly suggested a `deviceVerbose` note for this case, which wasn't added.
2. Ctrl-C now interrupting a running GPU computation at block granularity (#195) is a user-visible behavior change with no docs mention.

Per CLAUDE.md: "user-visible behavior changes should be documented in the appropriate `.rst` file unless purely internal" — a sentence or two in `gpu.rst` would cover both.

---

bwehlin (commented):
Adversarial review. I attacked all four fixes against the vendored Taskflow 4.0.1 source (exception anchoring/`_carry_out_promise` confirms `Future::get()` rethrows and `wait_for_all()` swallows — the #188 mechanism and pre-fix failure claim check out), verified all three #187 dispatch sites have a compiling CPU fall-through, confirmed #189 leaves no other shared mutable state in the manager, and confirmed the new test file is wired into the BUILD_CUDA_TESTER target. Three findings survived: a new error-path use-after-free hazard on pinned buffers when `.get()` rethrows past the flush loop (medium), a stop-request test that doesn't actually pin block-granularity interruption (low), and undocumented user-visible behavior changes required by CLAUDE.md (low).

---

PR #203, opened 2026-07-02
Align device-side rectangle iteration with the CPU path for custom integration bounds

### LINKED ISSUE #190: [cuda] Device-side rectangle iteration disagrees with the CPU path for non-default integration bounds [a, b] (latent)
## Summary

Two divergences between `include/sbear/cuda/cuda_pcf_kernel.cuh` and the CPU reference (`include/sbear/algorithms/functional/iterate_rectangles.hpp`) when integrating over a custom `[a, b]`:

1. **Start position** — `cuda_pcf_kernel.cuh:48-49` starts at `fi = gi = 0` (the first breakpoint), while the CPU does `max_time_iterator_prior_to` (`iterate_rectangles.hpp:44-45`) to begin at the last breakpoint ≤ `a`. For `a > 0` with breakpoints in `(0, a]`, the device path emits rectangles with wrong values (and negative widths for the leading segment).

2. **End clamping** — `cuda_pcf_kernel.cuh:89`: `t = max(fpts[fi].t, gpts[gi].t)` is not clamped to `b` (CPU: `std::min(std::max(...), b)`, `iterate_rectangles.hpp:95`), so the final rectangle over-integrates past a finite `b`.

## Impact

Currently **latent**: the Python bindings always pass `(0, max)`, so no user-visible wrong results today. But the public C++ factory API (`cuda_matrix_integrate_api.hpp`) explicitly accepts arbitrary `a`/`b`, and any caller using it gets GPU results that silently diverge from the CPU path — the worst kind of bug to debug later. Worth fixing (or asserting `a == 0 && b == inf` on the device path) before anything starts exercising these parameters.

bwehlin:
Two latent divergences between `cuda_pcf_kernel.cuh` and the CPU reference (`iterate_rectangles.hpp`) when integrating over non-default bounds `[a, b]`:

- The device walk started at the first breakpoint instead of the last breakpoint ≤ `a`, emitting wrong values (and a negative-width leading rectangle) whenever breakpoints exist in `(0, a]`. It now starts via a device-side binary search mirroring `max_time_iterator_prior_to`.
- The rectangle end was never clamped to `b`, over-integrating past a finite upper bound. It now clamps exactly like the CPU path.

Currently latent: the Python bindings always pass `(0, max)`, and behavior for those default bounds is unchanged. But the public C++ factory API accepts arbitrary `a`/`b`, and any caller using it got GPU results silently diverging from the CPU.

Fixes #190

## Tests

New `test/test_block_integrate_bounds.cu` (runs in the `BUILD_CUDA_TESTER` suite on the GPU runner): compares the CUDA block pdist against the CPU `integrate()` reference for breakpoints before `a`, a finite `b` inside the domain, both bounds custom, the default bounds, and a randomized 12-PCF sweep over three bound combinations — in float32 and float64.

## Verification

No GPU is available in the sandbox, so additionally the device function was extracted verbatim from the header (with `__device__` defined away) and fuzz-compared against `sb::iterate_rectangles` on 20,000 random PCF pairs and bounds: the fixed code matches the CPU output exactly on every case, while the previous code diverged on ~10,600 of them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -887,6 +887,7 @@ if (BUILD_TESTER)
         test/test_block_matrix_support.cu
         test/test_l1_dist.cu
         test/test_block_integrate.cu
+        test/test_block_integrate_bounds.cu
         test/test_norms.cu
         ${SB_TEST_SOURCES_COMMON}
     )
--- a/include/sbear/cuda/cuda_pcf_kernel.cuh
+++ b/include/sbear/cuda/cuda_pcf_kernel.cuh
@@ -31,8 +31,35 @@ namespace sb
       TriangleSkipMode skipMode;
     };
 
+    /// Index of the last breakpoint with t <= a (device analogue of the
+    /// host-side max_time_iterator_prior_to). Points are sorted by time.
+    template <typename Tt, typename Tv>
+    __device__ size_t cuda_max_time_index_prior_to(
+        const SimplePoint<Tt, Tv>* pts, size_t n, Tt a)
+    {
+      // upper_bound: first index with pts[i].t > a
+      size_t lo = 0;
+      size_t hi = n;
+      while (lo < hi)
+      {
+        size_t mid = lo + (hi - lo) / 2;
+        if (pts[mid].t <= a)
+        {
+          lo = mid + 1;
+        }
+        else
+        {
+          hi = mid;
+        }
+      }
+      return lo == 0 ? 0 : lo - 1;
+    }
+
     /// Walk two PCFs simultaneously through their breakpoints,
     /// calling cb(left, right, fValue, gValue) for each rectangle.
+    /// Mirrors the host-side iterate_rectangles (iterate_rectangles.hpp);
+    /// the two must stay in lockstep so CPU and GPU results agree for any
+    /// integration bounds [a, b].
     template<typename Tt, typename Tv, typename RectangleCallback>
     __device__ void cuda_pcf_iterate_rectangles(
         const SimplePoint<Tt, Tv>* rowPoints, size_t fOffset, size_t fsz,
@@ -45,11 +72,14 @@ namespace sb
       Tv fv;
       Tv gv;
 
-      size_t fi = 0;
-      size_t gi = 0;
-
       const SimplePoint<Tt, Tv>* fpts = rowPoints + fOffset;
       const SimplePoint<Tt, Tv>* gpts = colPoints + gOffset;
+
+      // Start at the last breakpoint <= a, like the CPU path -- starting at
+      // index 0 emits wrong values (and a negative-width leading rectangle)
+      // when there are breakpoints in (0, a].
+      size_t fi = cuda_max_time_index_prior_to(fpts, fsz, a);
+      size_t gi = cuda_max_time_index_prior_to(gpts, gsz, a);
 
       while (t < b)
       {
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -887,6 +887,7 @@ if (BUILD_TESTER)
         test/test_block_matrix_support.cu
         test/test_l1_dist.cu
         test/test_block_integrate.cu
+        test/test_block_integrate_bounds.cu
```
bwehlin, line 890: **No PR check ever compiles this file (or nvcc-compiles the fixed `.cuh`) before merge.**

Verified chain: `ci.yaml` (the `pull_request` workflow) calls `analysis.yaml` with `use_github_runners: true`, whose `cuda_test` job has `if: ${{ !inputs.use_github_runners }}` — so it is skipped on PRs. On GitHub-hosted runners `CMakeLists.txt` also forces `BUILD_CUDA_TESTER OFF` (`GITHUB_ACTIONS` set, no `GPUCOV_ENABLE`), so `sb_test` is built without any `.cu` sources. The first time `test_block_integrate_bounds.cu` meets a compiler is a manual `workflow_dispatch` of `analysis.yaml` or the `cuda_test` job inside `publish-pypi.yaml` — i.e., potentially at release time.

To de-risk this specific PR I host-compiled the new test file against the branch headers with `-DBUILD_WITH_CUDA` (the test body is pure host C++; `cuda_matrix_integrate_api.hpp` needs no nvcc): it compiles cleanly with the typed-test bodies instantiated for both `float` and `double`, and the harness was negative-controlled (injected errors are caught). So the latent-compile-error risk appears not to have materialized here — but consider dispatching `analysis.yaml` on this branch before merging so `cuda_test` actually runs the new suite on the GPU runner, since a *behavioral* GPU regression would otherwise also go undetected until release.

---

bwehlin (commented):
Adversarial review — I tried to break this fix; it held up on everything except one CI-visibility gap (inline comment).

What was attacked, and results:

- **Binary search vs `max_time_iterator_prior_to`**: compared line-by-line against `iterate_rectangles.hpp`. The device loop (`pts[mid].t <= a ? lo=mid+1 : hi=mid`, then `lo==0 ? 0 : lo-1`) is an exact `upper_bound`+`prev` analogue, including `a` exactly on a breakpoint, `a` before the first / after the last breakpoint, duplicate breakpoint times, and single-point PCFs.
- **b-clamping and termination**: rectangle straddling `b`, final rectangle ending exactly at `b`, `a == b` (zero rectangles emitted, no negative width), and the both-iterators-exhausted branch all match the CPU path.
- **Differential fuzz (independent reproduction of the PR's claim)**: I extracted the fixed device function verbatim (with `__device__` stubbed) and compared emitted rectangle sequences — not just integrals — against `sb::iterate_rectangles` on ~200k quantized-random cases per precision, deliberately including duplicate times, bounds on exact breakpoints, `a == b`, and `b = max`: **0 mismatches in float32 and float64**. The same harness with the pre-fix body (`fi = gi = 0`, no clamp) mismatches on 129,378/200,270 cases — consistent with the description's ~10,600/20,000 divergence claim.
- **"Default bounds unchanged" claim**: verified. With the `t = 0` construction invariant (first breakpoint pinned at 0), the search returns index 0 for `a = 0`, and the `min(..., b)` clamp is a no-op for `b = max`, so `(0, max)` behavior is bit-identical to the old code.
- **Fix-site completeness**: `cuda_pcf_iterate_rectangles` is the only device-side rectangle walk, and all six factory families (L1/Lp/L2 pdist and cdist) pass `a`/`b` through `pcf_block_op.cuh` unmodified into it, so the single fix site covers the whole public C++ surface. Minor note: the new tests exercise bounds only via the L1 pdist factory, but since the other factories share the fixed function and their `a`/`b` plumbing is pre-existing, that coverage is acceptable.
- **Test compilability despite no GPU CI**: see the inline comment — no PR check compiles the new `.cu`; I host-syntax-checked it (clean, both precisions instantiated) to close that gap for this PR.

No correctness bugs found in the fix itself.

---

Issue #25, opened 2026-06-07
[reductions] max_time segfaults on a size-0 reduced axis (empty inner dimension), while mean handles it


bwehlin:
_Found by an automated masspcf API bug scan (2026-06-07); independently reproduced and verified. **Severity: high.**_

### Summary
max_element in matrix_reduce.hpp builds a tmp vector of size inDimSize (==0 for an empty reduced dim) and then does 'auto init = f(*tmp.begin());' and 'std::transform_reduce(tmp.begin()+1, tmp.end(), ...)'. Dereferencing begin() on an empty vector is undefined behavior, hence the crash. mean uses reduce() which seeds with a default-constructed TPcf() so it survives the empty case.

### Minimal reproduction
```python
import masspcf as mpcf
A = mpcf.zeros((0,))
print("about to call max_time", flush=True)
r = mpcf.max_time(A, dim=0)   # SIGSEGV here (exit 139); same for max_time(zeros((3,0)), dim=1)
print("no crash:", r.shape, flush=True)
# Contrast (these return cleanly):
#   mpcf.mean(mpcf.zeros((0,)),  dim=0) -> shape (1,)
#   mpcf.mean(mpcf.zeros((3,0)), dim=1) -> shape (3,)
```

### Observed
mpcf.max_time over an empty reduced dimension crashes the interpreter with a Segmentation fault (core dumped), exit code 139 / signal SIGSEGV. Confirmed for both reported inputs: max_time(zeros((0,)), dim=0) and max_time(zeros((3,0)), dim=1). The minimal direct (non-subprocess) repro prints "about to call max_time" then dies with "Segmentation fault (core dumped)", EXIT=139 -- a hard crash with no Python traceback. By contrast, mean over the identical empty reductions returns cleanly: mean(zeros((0,)), dim=0) -> Shape(1) and mean(zeros((3,0)), dim=1) -> Shape(3).

### Expected
A valid Python call must never segfault. max_time should either return a well-defined result for an empty reduction (e.g. 0.0, consistent with the unary fallback that already maps empty PCFs to 0; or NaN as numpy's empty-max would raise/produce) or raise a clean Python exception (e.g. ValueError for zero-size reduce). Its empty-axis behavior should be consistent with mean, which returns a valid (NaN-valued) reduced tensor without crashing.

### Root-cause hint
[...]

PR #204, opened 2026-07-02
Fix size_t underflow in subdivide() for empty ranges

### LINKED ISSUE #186: [core] subdivide() underflows for nItems == 0 — parallel_reduce/average on an empty range reads ~2^64 elements out of bounds
## Summary

`include/sbear/algorithms/subdivide.hpp:24` (block-clamping branch): with `nItems == 0`, the first block emplaced is `(0, min(blockSize-1, 0)) = (0, 0)`; then `back().second >= nItems` is true and `back().second = nItems - 1` **wraps to SIZE_MAX**, returning `{(0, SIZE_MAX)}`.

Consumed by `parallel_reduce` (`include/sbear/algorithms/functional/reduce.hpp:215-221`), each block task runs:

```cpp
for (auto i = block.first; i <= block.second; ++i)
  accumulators[iBlock] += *(begin + i);
```

i.e. on an empty range it dereferences `begin + 0 … begin + SIZE_MAX` — immediate out-of-bounds read / segfault.

## Reachability

- `sb::average(fs)` with an empty vector (`include/sbear/functional/pcf.hpp:325-331`) — which additionally divides by `fs.size() == 0`.
- Directly from Python via the bindings: `cpp.Pcf_*.average([])` / `cpp.Pcf_*.parallel_reduce([], cb)` (`src/python/functional/py_pcf.cpp:58-70`) — no empty guard anywhere in the chain.

`test/test_subdivide.cpp` never tests `nItems == 0`.

Related history: the empty-input crash family was fixed for `max_time` (#25, #46) but this entry point was missed.

## Suggested fix

Return an empty boundary list from `subdivide()` when `nItems == 0`, and give `parallel_reduce`/`average` a clean empty-range error (like the one `max_element` got in `matrix_reduce.hpp`).

bwehlin:
`subdivide(bs, 0)` clamped the first block's end to `nItems - 1`, which wraps to `SIZE_MAX`, so `parallel_reduce`/`average` on an empty range (reachable from Python via `cpp.Pcf_*.average([])` / `parallel_reduce([], cb)`) read ~2^64 elements out of bounds.

- `subdivide` now returns no blocks for zero items.
- `parallel_reduce` rejects an empty range with `std::invalid_argument` (a Python `ValueError`), mirroring the empty-dimension error `max_element` got for #25/#46 — a pairwise combine has no identity element. This also fixes `average`'s division by `fs.size() == 0` on the same path.

New regression tests: `Subdivide.NoItems`, `ParallelReduce.EmptyRangeThrows`, `ParallelReduce.AverageOfEmptyVectorThrows`; the `subdivide` example table in `docs/internals/cuda_engine.rst` documents the empty case.

Fixes #186

---

### DIFF
--- a/include/sbear/algorithms/functional/reduce.hpp
+++ b/include/sbear/algorithms/functional/reduce.hpp
@@ -10,6 +10,7 @@
 #include <numeric>
 #include <algorithm>
 #include <iterator>
+#include <stdexcept>
 
 #include <taskflow/taskflow.hpp>
 #include <taskflow/algorithm/reduce.hpp>
@@ -197,6 +198,14 @@ namespace sb
 
     auto chunksz = 2;
     auto sz = std::distance(begin, end);
+    if (sz == 0)
+    {
+      // A pairwise combine has no identity element, so an empty range cannot
+      // be reduced. Reject up front with a clean, catchable exception
+      // (pybind11 maps std::invalid_argument to a Python ValueError) --
+      // mirrors the empty-dimension error in matrix_reduce.hpp.
+      throw std::invalid_argument("Cannot reduce an empty range of PCFs");
+    }
     auto blocks = subdivide(chunkszFirst , sz);
 
     auto nAll = std::accumulate(begin, end, static_cast<size_t>(0ul), [](size_t n, const pcf_type& f){ return f.points().size() + n; });
--- a/test/test_parallel_reduce.cpp
+++ b/test/test_parallel_reduce.cpp
@@ -17,3 +17,22 @@ TEST(ParallelReduce, AddThreeFunctions)
 
   EXPECT_EQ(res, (pcfs[0] + pcfs[1] + pcfs[2]));
 }
+
+// Regression for issue #186: an empty range used to reach subdivide's
+// SIZE_MAX-wrapped block and read far out of bounds.
+TEST(ParallelReduce, EmptyRangeThrows)
+{
+  std::vector<sb::Pcf_f64> pcfs;
+
+  auto op = [](const typename sb::Pcf_f64::rectangle_type& rect) {
+    return rect.f_value + rect.g_value;
+  };
+
+  EXPECT_THROW(sb::parallel_reduce(pcfs.begin(), pcfs.end(), op), std::invalid_argument);
+}
+
+TEST(ParallelReduce, AverageOfEmptyVectorThrows)
+{
+  std::vector<sb::Pcf_f64> pcfs;
+  EXPECT_THROW(sb::average(pcfs), std::invalid_argument);
+}
--- a/docs/internals/cuda_engine.rst
+++ b/docs/internals/cuda_engine.rst
@@ -93,6 +93,7 @@ The square root targets roughly square blocks, which is a good default for 2D CU
    subdivide(3, 10)  ->  [0,2], [3,5], [6,8], [9,9]
    subdivide(4, 8)   ->  [0,3], [4,7]
    subdivide(5, 3)   ->  [0,2]
+   subdivide(5, 0)   ->  (no bands)
 
 Row bands and column bands are computed independently: ``rowBands = subdivide(blockSide, nRows)``, ``colBands = subdivide(blockSide, nCols)``. This means rectangular matrices are handled naturally -- the row and column band counts can differ.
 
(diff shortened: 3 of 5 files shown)


include/sbear/algorithms/functional/reduce.hpp
```diff
@@ -197,6 +198,14 @@ namespace sb
...
     auto chunksz = 2;
     auto sz = std::distance(begin, end);
+    if (sz == 0)
+    {
+      // A pairwise combine has no identity element, so an empty range cannot
+      // be reduced. Reject up front with a clean, catchable exception
+      // (pybind11 maps std::invalid_argument to a Python ValueError) --
+      // mirrors the empty-dimension error in matrix_reduce.hpp.
+      throw std::invalid_argument("Cannot reduce an empty range of PCFs");
+    }
```
bwehlin, line 208: **[Low] Empty-range semantics now diverge from the serial `sb::reduce` in this same header.**

`sb::reduce` (line 58, ~140 lines up) reduces an empty range to `TPcf()` — the zero function — via `std::reduce(begin, end, TPcf(), ...)`. After this change the two reduce entry points disagree on empty input: serial silently returns the zero PCF, parallel throws. The comment's justification ("a pairwise combine has no identity element") is undercut by serial `reduce` using `TPcf()` as exactly that init element in the same file.

Same pattern one layer up: `parallel_tensor_reduce` in `matrix_reduce.hpp` funnels per-slice reduction through the serial `reduce(first, last, op)`, so an empty slice (if a zero-sized dim is constructible) yields the zero function and `mean` then divides by `inDimSize == 0` — silently producing inf/nan rather than the ValueError this path now raises.

Not a bug introduced here — throwing matches the `max_element` precedent and is strictly better than the old OOB read — but consider either giving `sb::reduce` the same guard (or documenting why it keeps zero-identity semantics), so callers don't get crash-vs-zero behavior depending on which overload they pick.

---

test/test_parallel_reduce.cpp
```diff
@@ -17,3 +17,22 @@ TEST(ParallelReduce, AddThreeFunctions)
...
+    return rect.f_value + rect.g_value;
+  };
+
+  EXPECT_THROW(sb::parallel_reduce(pcfs.begin(), pcfs.end(), op), std::invalid_argument);
+}
+
+TEST(ParallelReduce, AverageOfEmptyVectorThrows)
+{
+  std::vector<sb::Pcf_f64> pcfs;
+  EXPECT_THROW(sb::average(pcfs), std::invalid_argument);
```
bwehlin, line 37: **[Low] The "Python `ValueError`" claim is untested — no Python-side regression test.**

The PR description advertises the Python-visible contract (`ValueError` from the bindings), and the issue explicitly lists the Python entry points as the reachable surface, but the new tests only pin `std::invalid_argument` at the C++ layer. Nothing in `test/python/` exercises `average`/`parallel_reduce` at all (the bindings currently have zero Python test coverage), so the pybind translation step — the part users actually see — is unpinned. A two-liner would close it:

```python
    cpp.Backend_f64_f64.average([])
```

Nit: the PR description and code comment say `cpp.Pcf_*.average([])`, but the binding is a static on `Backend_*` (`src/python/functional/py_pcf.cpp` line 210), not on `Pcf_*`.

---

bwehlin (commented):
Adversarial review. Attacked: (1) all `subdivide()` callers now that it can return zero blocks — `CudaBlockScheduler::compute_blocks` guards `nRows/nCols == 0` itself, and both call sites in `parallel_reduce` are behind the new size guard / a `size() > 1` loop condition, so no caller sees an unexpected empty block list; (2) the fix itself — compiled the PR-branch header standalone with ASan/UBSan and cross-checked `subdivide(bs, n)` against a reference implementation for bs 1–12, n 0–100: empty case correct, no off-by-one regression; (3) exception plumbing — the throw happens on the calling thread before any taskflow construction, no custom pybind exception translators exist in `src/python/`, so the default `std::invalid_argument` → `ValueError` mapping claim holds; (4) tests — both files are in the `sb_test` source list in CMakeLists.txt; (5) the `docs/internals/cuda_engine.rst` table row matches verified behavior.

Two low-severity findings posted inline: the empty-range semantics now diverge from the serial `sb::reduce` (and the tensor `mean` path) in the same headers, and the advertised Python `ValueError` contract has no Python-side test. Minor additional nit: `average([])` surfaces the message "Cannot reduce an empty range of PCFs", which names an operation the caller didn't invoke.

---

PR #205, opened 2026-07-02
Return an empty result when evaluating a PCF tensor at an empty grid

### LINKED ISSUE #192: [eval] Evaluating a PCF tensor at an empty time array is undefined behavior (domain_indices[0] on empty vector)
## Summary

`include/sbear/algorithms/tensor_eval.hpp:69` — `combined_idx.resize(elem_idx.size() + domain_indices[0].size())` indexes `domain_indices[0]` without checking that the domain is non-empty. With a 0-point domain tensor, `domain_indices` is empty and `[0]` is UB (garbage resize size or crash), executed once per tensor element.

## Reachability from Python

```python
T = sb.zeros((3,))          # PcfTensor
T(np.array([]))             # empty evaluation grid
```

goes through `py_tensor.hpp:467-475` with no size guard, and `_tensor_base.py:869-871` adds none either. NumPy convention for the analogous case (`f(np.array([]))` elementwise) is to return an empty result, not crash.

## Suggested fix

Early-return an empty result tensor of shape `elems.shape() + (0,)` when the domain has zero points (or raise a clean `ValueError` if empty grids are considered invalid). Found independently by two review passes.

bwehlin:
`tensor_eval`'s grid overload indexed `domain_indices[0]` without checking that the domain has any points, so `T(np.array([]))` was undefined behavior (executed once per tensor element). It now early-returns: the output already has shape `elems.shape() + domain.shape()`, i.e. zero entries — matching numpy's convention that an empty evaluation grid yields an empty result.

New Python regression tests cover the empty numpy-array, empty-list, and empty-`FloatTensor` inputs across all PCF dtypes (result shapes `(2, 0)` / `(2, 2, 0)`).

Fixes #192

---

### DIFF
--- a/include/sbear/algorithms/tensor_eval.hpp
+++ b/include/sbear/algorithms/tensor_eval.hpp
@@ -41,6 +41,13 @@ namespace sb
     std::vector<DomainT> domain_values;
     std::vector<std::vector<size_t>> domain_indices;
     auto n = domain.size();
+    if (n == 0)
+    {
+      // out already has shape elems.shape() + domain.shape(), i.e. zero
+      // entries -- nothing to evaluate. Bail before eval_elem indexes
+      // domain_indices[0] (numpy convention: empty grid -> empty result).
+      return;
+    }
     domain_values.reserve(n);
     domain_indices.reserve(n);
     walk(domain, [&](const std::vector<size_t>& idx) {
--- a/test/python/test_pcf_tensor_eval.py
+++ b/test/python/test_pcf_tensor_eval.py
@@ -130,6 +130,29 @@ class TestPcfTensorEvalArray:
         assert result.shape == (2, 2)
         npt.assert_array_almost_equal(result, [[1, 5], [3, 1]])
 
+    def test_empty_times_array(self, dtypes):
+        # Regression for issue #192: an empty evaluation grid used to hit
+        # undefined behavior instead of returning an empty result.
+        pcf_dtype, np_dtype = dtypes
+        X = make_1d_tensor(pcf_dtype, np_dtype)
+        result = X(np.array([], dtype=np_dtype))
+        assert isinstance(result, np.ndarray)
+        assert result.shape == (2, 0)
+
+    def test_empty_times_list(self, dtypes):
+        pcf_dtype, np_dtype = dtypes
+        X = make_2d_tensor(pcf_dtype, np_dtype)
+        result = X([])
+        assert result.shape == (2, 2, 0)
+
+    @pytest.mark.parametrize("float_dtype", [np.float32, np.float64])
+    def test_empty_float_tensor_input(self, dtypes, float_dtype):
+        pcf_dtype, np_dtype = dtypes
+        X = make_1d_tensor(pcf_dtype, np_dtype)
+        t = sb.FloatTensor(np.array([], dtype=float_dtype))
+        result = X(t)
+        assert result.shape == (2, 0)
+
 
 class TestPcfTensorEvalParallel:
     """Verify that parallel tensor_eval produces correct results.


include/sbear/algorithms/tensor_eval.hpp
```diff
@@ -41,6 +41,13 @@ namespace sb
     std::vector<DomainT> domain_values;
     std::vector<std::vector<size_t>> domain_indices;
     auto n = domain.size();
+    if (n == 0)
+    {
+      // out already has shape elems.shape() + domain.shape(), i.e. zero
+      // entries -- nothing to evaluate. Bail before eval_elem indexes
+      // domain_indices[0] (numpy convention: empty grid -> empty result).
+      return;
+    }
```
bwehlin, line 50: The comment's invariant — "out already has shape elems.shape() + domain.shape(), i.e. zero entries" — is **false for a rank-0 domain tensor**, and this guard now fires for that case too.

`Tensor<T>::size()` special-cases rank-0 to return 0 (`include/sbear/tensor.tpp:1187-1195`), even though `walk_impl` explicitly treats a rank-0 tensor as holding **one** element ("A rank-0 (shape.empty()) tensor holds exactly one element and is visited once", `include/sbear/walk.hpp`). So for a shape-`()` domain, `n == 0` here, but `out` has shape `elems.shape() + ()` = `elems.shape()` — a full-size, non-empty buffer that is now returned untouched (all default-initialized values).

Reachable from Python:

```python
X = make_1d_tensor(...)            # [f, g]
X(sb.FloatTensor(np.array(2.0)))   # rank-0 tensor -> silently returns [0, 0] instead of [f(2), g(2)]
```

(`FloatTensor(np.array(2.0))` builds a shape-`()` `sb::Tensor` via `ndarray_to_tensor_64`, and `FunctionTensorMixin.__call__` routes it to the `const sb::Tensor<Tt>&` overload in `src/python/py_tensor.hpp:477-481`. The numpy 0-d path is unaffected because `NumpyTensor::size()` returns `m_arr.size()` = 1.)

To be fair, this is not a regression: pre-PR, `n == 0` made `order`/`sorted_domain` empty and the per-element copy loop run zero times, so the same silent zeros came out (the `domain_indices[0]` access happened to be valid because `walk` had pushed one entry). But this PR is adding the guard that decides what "empty domain" means, so it's the right place to get it right: either key the early return on an actual zero extent (`std::any_of(shape, [](auto s){ return s == 0; })`, matching `walk_impl`'s emptiness test) and let rank-0 evaluate its single point, or fix the `Tensor::size()` rank-0 inconsistency — and at minimum correct this comment, which currently documents an invariant the code doesn't have.

---

test/python/test_pcf_tensor_eval.py
```diff
@@ -130,6 +130,29 @@ def test_float_tensor_input(self, dtypes, float_dtype):
...
         npt.assert_array_almost_equal(result, [[1, 5], [3, 1]])
 
+    def test_empty_times_array(self, dtypes):
+        # Regression for issue #192: an empty evaluation grid used to hit
+        # undefined behavior instead of returning an empty result.
+        pcf_dtype, np_dtype = dtypes
+        X = make_1d_tensor(pcf_dtype, np_dtype)
+        result = X(np.array([], dtype=np_dtype))
+        assert isinstance(result, np.ndarray)
+        assert result.shape == (2, 0)
```
bwehlin, line 140: Two coverage gaps in the new tests:

1. **Every new test uses a 1-D empty grid.** The `(2, 2, 0)` in the PR description comes from a 2×2 *elements* tensor with a 1-D empty grid, not from a multidimensional *domain*. The fix itself is shape-agnostic (`domain.size() == 0` catches e.g. `np.empty((2, 0))` → result shape `(2, 2, 0)` with the trailing dims from the domain), but nothing exercises that path, even though `test_1d_tensor_2d_times` above establishes multidim grids as supported input. One case with a multidim empty grid (e.g. `X(np.empty((2, 0), dtype=np_dtype))` → shape `(2, 2, 0)`) would pin down that `eval_out_shape` and the early return agree for rank>1 empty domains.

2. **No dtype assertion.** The sibling tests (`test_1d`, `test_1d_tensor_1d_times`) assert `result.dtype == np_dtype`; the empty-grid tests only check shape. Since the early return depends entirely on the binding having pre-allocated the output correctly (`py::array_t<Tv>` in `py_tensor.hpp`), asserting the dtype of the empty result is exactly the metadata this code path is responsible for, and it's one line.

---

bwehlin (commented):
Adversarial review of the empty-grid fix. Attack surface checked: guard placement (nothing before it touches `domain_indices`), the other `tensor_eval` overload (scalar `x` — no domain indexing, safe), the single-`Pcf` array `__call__` in `py_pcf.cpp` (all loops bounded by `n`, safe at `n == 0`), empty *elements* tensor vs. non-empty grid (walk skips, unchanged), all three Python `__call__` bindings (numpy, list, `sb::Tensor` — the C++-level guard covers them all, including the CUDA-built module since the header is shared), and output construction for zero-extent shapes (`py::array_t` and `sb::Tensor` both handle it).

Two findings survived, posted inline:

1. `Tensor::size()` returns 0 for rank-0 tensors while `walk` visits them once, so the new guard also fires for a scalar (shape-`()`) domain tensor, where the output is *not* empty — it's returned full of default-initialized values. Same silent zeros came out pre-PR, so not a regression, but the new comment documents an invariant ("zero entries") that doesn't hold for that reachable case.
2. Test gaps: no multidimensional empty-grid case, and the empty tests drop the `result.dtype` assertion their siblings make.

---

PR #206, opened 2026-07-02
Validate from_serial_content shapes and use signed stride arithmetic

### LINKED ISSUE #193: [io] from_serial_content: trailing dimensions never validated — content shape (N,1) or enumeration last-dim != 2 reads out of bounds
## Summary

Follow-up to #37 (which fixed the unchecked *enumeration index values*): `src/python/functional/py_make_from_serial_content.cpp:35-49` validates `content.ndim == 2` and `enumeration.ndim >= 2`, but never checks the trailing extents:

- `content.shape(1) >= 2` is assumed — `get_element(content, {i, 1})` (line ~82) does raw unchecked pointer arithmetic (`py_np_support.hpp:31-36`), so `content` of shape `(N, 1)` reads one element past each row.
- `enumeration.shape(-1) == 2` is assumed — the comment at line 46 just states "Last dim is always 2 for [start, end)"; an enumeration with last dim 1 reads `stop` out of bounds.

## Failure scenario

```python
sb.from_serial_content(content=np.zeros((5, 1)), enumeration=...)   # heap over-read per row
sb.from_serial_content(content=..., enumeration=np.zeros((3, 1), dtype=np.int64))
```

Garbage PCFs (silent) or a crash, instead of a clean shape error. The Python wrapper (`stablebear/serialize.py`) adds no shape validation either.

## Suggested fix

Add `content_buf.shape[1] == 2` and `enumeration_buf.shape[ndim-1] == 2` checks next to the existing ndim checks, with the same friendly error style.

### LINKED ISSUE #194: [bindings] Unsigned stride arithmetic in py_np_support get_element/strides breaks negative-stride numpy inputs
## Summary

Same bug class as the fixed #23, in a different spot: `src/python/py_np_support.hpp:31-36`:

```cpp
auto offset = std::inner_product(idx.begin(), idx.end(), arr.strides(), 0_uz);
offset /= arr.itemsize();
```

accumulates `idx · strides` with an **unsigned** initial value (`0_uz`, `include/sbear/config.hpp:61`) and divides in unsigned. NumPy strides are signed; a negative stride (e.g. from `content[::-1]`) wraps to ~2^64 and the unsigned division then produces a garbage offset → out-of-bounds read.

Affected users of this helper:
- `py_make_from_serial_content.cpp` (~lines 55-56, 82) — `from_serial_content` with a reversed/negatively-strided `content` or `enumeration` array reads garbage or crashes.
- `NumpyTensor::strides()` (`py_np_support.hpp:58-66`) also casts negative strides to `size_t`.

(For contrast, `py_np_tensor_convert.cpp:25` uses a signed `0_z` accumulator and is correct — the fix is to do the same here, deferring the itemsize division or using signed throughout.)

## Failure scenario

```python
content = np.ascontiguousarray(big_content)[::-1]   # negative stride, no copy
sb.from_serial_content(content, enumeration)         # garbage PCFs or segfault
```

## Suggested fix

Use `pybind11::ssize_t`/`0_z` for the accumulation and division (matching `py_np_tensor_convert.cpp`), or force a contiguous copy on entry for negatively-strided inputs.

bwehlin:
Two OOB-read classes in the serial-content path, grouped because they hit the same function and share regression tests:

- **#193** — trailing dimensions were never validated: `content` of shape `(N, 1)` read one element past every row, and an `enumeration` whose last dim ≠ 2 read `stop` out of bounds. Both now raise a clean shape error in the style of the existing ndim checks.
- **#194** — `get_element` and the enumeration base offset accumulated `idx · strides` with an unsigned initial value and divided in unsigned; numpy strides are signed, so a reversed (negative-stride) view wrapped to ~2^64 and produced garbage offsets. Accumulate and divide as signed, matching the already-correct `py_np_tensor_convert.cpp`. `NumpyTensor::strides()` is now signed too, like `sb::Tensor::strides()` (it had no consumers relying on the unsigned cast).

New Python regression tests: wrong trailing dims for both arrays raise, and reversed (negative-stride) `content`/`enumeration` views read the right elements.

Fixes #193, fixes #194

---

### DIFF
--- a/src/python/functional/py_make_from_serial_content.cpp
+++ b/src/python/functional/py_make_from_serial_content.cpp
@@ -37,9 +37,19 @@ namespace sb_py
       throw std::runtime_error("content should have 2 dimensions (content has shape " + shape_to_string(content) + ").");
     }
 
+    if (content_buf.shape[1] != 2)
+    {
+      throw std::runtime_error("content should have 2 columns, [time, value] (content has shape " + shape_to_string(content) + ").");
+    }
+
     if (enumeration_buf.ndim < 2)
     {
       throw std::runtime_error("enumeration must have at least 2 dimensions");
+    }
+
+    if (enumeration_buf.shape[enumeration_buf.ndim - 1] != 2)
+    {
+      throw std::runtime_error("enumeration's last dimension should have size 2, [start, stop) (enumeration has shape " + shape_to_string(enumeration) + ").");
     }
 
     std::vector<size_t> targetShape(enumeration_buf.ndim - 1);
@@ -52,7 +62,9 @@ namespace sb_py
 
     sb::walk(target, [&target, &content, &enumeration](const std::vector<size_t>& idx) {
 
-      auto enumerationBaseOffset = std::inner_product(idx.begin(), idx.end(), enumeration.strides(), 0_uz);
+      // Signed accumulation: strides are negative for reversed views, and an
+      // unsigned division of the wrapped sum would produce a garbage offset.
+      auto enumerationBaseOffset = std::inner_product(idx.begin(), idx.end(), enumeration.strides(), py::ssize_t{0});
       enumerationBaseOffset /= enumeration.itemsize();
 
       auto* enumerationBuf = enumeration.unchecked().data();
--- a/src/python/py_np_support.hpp
+++ b/src/python/py_np_support.hpp
@@ -30,7 +30,9 @@ std::string shape_to_string(pybind11::array_t<T> arr)
 template <typename T>
 T& get_element(pybind11::array_t<T>& arr, const std::vector<pybind11::ssize_t>& idx)
 {
-  auto offset = std::inner_product(idx.begin(), idx.end(), arr.strides(), 0_uz);
+  // NumPy strides are signed (negative for reversed views), so the offset
+  // must accumulate and divide as a signed quantity.
+  auto offset = std::inner_product(idx.begin(), idx.end(), arr.strides(), pybind11::ssize_t{0});
   offset /= arr.itemsize();
   return *(static_cast<T*>(arr.request().ptr) + offset);
 }
@@ -55,12 +57,14 @@ public:
     return m_arr.shape(i);
   }
 
-  [[nodiscard]] std::vector<size_t> strides() const
+  [[nodiscard]] std::vector<pybind11::ssize_t> strides() const
   {
-    std::vector<size_t> s;
+    // Signed, like sb::Tensor::strides() -- numpy strides are negative for
+    // reversed views and must not be wrapped through size_t.
+    std::vector<pybind11::ssize_t> s;
     s.resize(m_arr.ndim());
     std::transform(m_arr.strides(), m_arr.strides() + m_arr.ndim(), s.begin(), [this](pybind11::ssize_t n) {
-      return static_cast<size_t>(n / m_arr.itemsize());
+      return n / m_arr.itemsize();
     });
     return s;
   }
--- a/test/python/test_from_serial_content.py
+++ b/test/python/test_from_serial_content.py
@@ -118,6 +118,64 @@ def test_fsc_negative_start_raises(dtype):
         sb.from_serial_content(content, enumeration)
 
 
+def test_fsc_content_wrong_trailing_dim_raises():
+    """content with anything but 2 columns must raise, not read past each row.
+
+    Regression for #193: content of shape (N, 1) read one element past every
+    row (heap over-read) instead of raising a shape error.
+    """
+    enumeration = np.array([[0, 2]], dtype=np.int64)
+    with pytest.raises(RuntimeError):
+        sb.from_serial_content(np.zeros((5, 1)), enumeration)
+    with pytest.raises(RuntimeError):
+        sb.from_serial_content(np.zeros((5, 3)), enumeration)
+
+
+def test_fsc_enumeration_wrong_trailing_dim_raises():
+    """enumeration whose last dim is not 2 must raise, not read stop OOB.
+
+    Regression for #193.
+    """
+    content = np.array([[0.0, 1.0], [2.0, 3.0]])
+    with pytest.raises(RuntimeError):
+        sb.from_serial_content(content, np.zeros((3, 1), dtype=np.int64))
+    with pytest.raises(RuntimeError):
+        sb.from_serial_content(content, np.zeros((3, 3), dtype=np.int64))
+
+
+def test_fsc_negative_stride_content():
+    """A reversed (negative-stride) content view must read the right rows.
+
+    Regression for #194: unsigned stride arithmetic wrapped the negative
+    stride to ~2^64, producing garbage offsets (OOB reads).
+    """
+    content = np.array([[0.0, 1.0], [0.0, 2.0], [0.0, 3.0], [0.0, 4.0]])
+    rev = content[::-1]
+    assert rev.strides[0] < 0
+    enumeration = np.array([[0, 1], [1, 2], [2, 3], [3, 4]], dtype=np.int64)
+
+    X = sb.from_serial_content(rev, enumeration)
+
+    for i in range(4):
+        assert X[i] == sb.Pcf(np.array([[0.0, 4.0 - i]]))
+
+
+def test_fsc_negative_stride_enumeration():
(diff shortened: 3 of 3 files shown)


test/python/test_from_serial_content.py
```diff
@@ -118,6 +118,64 @@ def test_fsc_negative_start_raises(dtype):
...
+
+
+def test_fsc_negative_stride_content():
+    """A reversed (negative-stride) content view must read the right rows.
+
+    Regression for #194: unsigned stride arithmetic wrapped the negative
+    stride to ~2^64, producing garbage offsets (OOB reads).
+    """
+    content = np.array([[0.0, 1.0], [0.0, 2.0], [0.0, 3.0], [0.0, 4.0]])
+    rev = content[::-1]
```
bwehlin, line 153: **[medium] These negative-stride regression tests almost certainly pass on pre-fix `main` too, so they don't actually pin #194.**

The old unsigned arithmetic coincidentally lands on the correct address on all mainstream targets. Mechanism: for any slice-derived view, the accumulated byte offset is a multiple of `itemsize`, so the unsigned division of the wrapped sum is *exact*: for stride `-16`, `idx=1`, `itemsize=8`: `(2^64 - 16)/8 = 2^61 - 2`. Then `ptr + offset` multiplies back by `sizeof(T) == itemsize`, giving a byte offset of `(2^61-2)*8 ≡ -16 (mod 2^64)` — pointer arithmetic wraps back to the correct element.

I verified this empirically with a pybind11 module replicating `main`'s exact `get_element` expression (`std::inner_product(..., std::size_t{0})`, then `offset /= arr.itemsize()`), GCC 13, x86-64, at both `-O0` and `-O2`, on a real `content[::-1]` numpy view (which does pass into `py::array_t<double>` uncopied — I checked the data pointer):

```
rev[0,1]: main=4.0 pr=4.0 expected=4.0
rev[1,1]: main=3.0 pr=3.0 expected=3.0   (unsigned offset was 2305843009213693951)
rev[2,1]: main=2.0 pr=2.0 expected=2.0
rev[3,1]: main=1.0 pr=1.0 expected=1.0
```

Same cancellation applies to `enumerationBaseOffset` (`sizeof(long long) == itemsize == 8`).

So the old code was UB (unsigned wrap is fine, but `ptr + 2^61` is pointer-arithmetic overflow) rather than an observable "garbage offsets (OOB reads)" — the fix is still worth having as UB hygiene, but:

1. The docstrings here, the PR description, and the code comments overstate what happened ("producing garbage offsets (OOB reads)") — suggest rewording to "undefined behavior (pointer-arithmetic overflow); happens to compute the right address on current compilers".
2. As regression tests these are green both before and after the change; the only way to actually catch a regression of this class is UBSan (`-fsanitize=pointer-overflow`). Worth either noting that in the docstrings or adding a UBSan job if you want this pinned.

(Suggest double-checking by running these two tests against a `main` build — I could not do a full build in the review environment, but the standalone replication is faithful to the types and expressions involved.)

---

src/python/functional/py_make_from_serial_content.cpp
```diff
@@ -37,11 +37,21 @@ namespace sb_py
...
+
     if (enumeration_buf.ndim < 2)
     {
       throw std::runtime_error("enumeration must have at least 2 dimensions");
     }
 
+    if (enumeration_buf.shape[enumeration_buf.ndim - 1] != 2)
+    {
+      throw std::runtime_error("enumeration's last dimension should have size 2, [start, stop) (enumeration has shape " + shape_to_string(enumeration) + ").");
+    }
```
bwehlin, line 53: **[low] Unstated backward-compat break: only trailing dims `< 2` were OOB; `> 2` previously worked correctly and is now rejected.**

`get_element` reads columns 0 and 1 via strides, so `content` of shape `(N, 3)` (e.g. `[time, value, weight]` data passed directly) and an `enumeration` with last dim `> 2` were read fully in-bounds before this PR and produced correct results — the OOB read only existed for trailing extents `< 2`. After this PR those inputs raise.

Enforcing the documented `(m, 2)` / `(..., 2)` contract is defensible (and matches the suggestion in #193), but the PR description frames every newly rejected shape as a previous heap over-read, which is only true for the `< 2` cases, and `test_fsc_content_wrong_trailing_dim_raises` / `test_fsc_enumeration_wrong_trailing_dim_raises` now enshrine the rejection of `(5, 3)` / `(3, 3)`. If anyone was passing wider arrays and silently getting correct PCFs from the first two columns, this is a silent-success → error change — worth a line in the PR description/changelog confirming the strictness is intended, rather than only validating `>= 2`.

---

src/python/py_np_support.hpp
```diff
@@ -30,7 +30,9 @@ std::string shape_to_string(pybind11::array_t<T> arr)
 template <typename T>
 T& get_element(pybind11::array_t<T>& arr, const std::vector<pybind11::ssize_t>& idx)
 {
-  auto offset = std::inner_product(idx.begin(), idx.end(), arr.strides(), 0_uz);
+  // NumPy strides are signed (negative for reversed views), so the offset
+  // must accumulate and divide as a signed quantity.
+  auto offset = std::inner_product(idx.begin(), idx.end(), arr.strides(), pybind11::ssize_t{0});
```
bwehlin, line 35: **[low, pre-existing but adjacent] Sum-then-divide is still wrong for strides that aren't multiples of `itemsize` — the new comment overclaims strided-view correctness.**

The signed fix makes slice-derived views correct, but the addressing scheme (accumulate byte offsets, then divide once by `itemsize`; and the per-stride `n / m_arr.itemsize()` in `strides()` below) silently mis-addresses any array whose byte strides aren't multiples of `itemsize`. Structured-array field views are a real, user-reachable case: `py::array_t<T>` with only `forcecast` does **not** force contiguity or copy an equivalent-dtype input, so such a view arrives uncopied. Verified against this exact (post-fix, signed) expression:

```python
rec = np.zeros((3, 2), dtype=[('t', 'f8'), ('pad', 'f4')])
fv = rec['t']            # float64 view, strides (24, 12), itemsize 8 — passes through uncopied
```

`get_element(fv, {i, 1})` computes `(24*i + 12) / 8 = 3*i + 1` (truncated) and reads byte `24*i + 8` instead of `24*i + 12` — silently wrong values (`0.0` instead of `1.0 / 11.0 / 21.0` in my test), no error. `np.lib.stride_tricks.as_strided` inputs hit the same thing.

This predates the PR, so no need to fix here — but since the new comments assert "the offset must accumulate and divide as a signed quantity" as the correctness story, consider either (a) doing byte-based addressing (`reinterpret_cast<char*>(ptr) + byte_offset`) so the division disappears, or (b) rejecting/copying inputs with `stride % itemsize != 0`, or at least tempering the comment.

---

bwehlin (commented):
Adversarial review of the #193/#194 fix. Attack surface covered: completeness of the new shape checks (0-column/empty/ndim edge cases — all handled), all consumers of `NumpyTensor::strides()` after the unsigned→signed change (verified none exist outside the `IsTensor` concept, which only requires `Iterable` — the PR's "no consumers" claim holds), signedness of the remaining arithmetic on the path (`lastStride`, mixed `size_t`-index × signed-stride accumulation — correct under C++20 modular conversion), signed-division truncation semantics, and whether the new tests actually reach the fixed code (verified with a real pybind11 module: negative-stride float64/int64 views pass into `py::array_t` uncopied, so they do).

Three findings, none blocking:

1. **[medium]** The negative-stride regression tests almost certainly pass on pre-fix `main`: the unsigned wrap cancels through the exact division and pointer arithmetic on all mainstream targets (verified empirically with a replication of `main`'s exact expressions). The fix removes UB, but the tests don't pin it and the "garbage offsets (OOB reads)" claims are overstated for #194.
2. **[low]** Trailing dims `> 2` (`content (N,3)`, enumeration last dim 3) previously produced correct in-bounds results and are now rejected — a legitimate strictness choice, but an unstated silent-success → error compat change; only `< 2` was OOB.
3. **[low, pre-existing]** The sum-then-divide addressing is still silently wrong for strides that aren't `itemsize` multiples (structured-array field views pass through uncopied and read wrong elements — verified); the new comments overclaim strided-view correctness.

The core fix itself is correct: the #193 checks close a real heap over-read for trailing dims `< 2`, and the signed accumulation matches `py_np_tensor_convert.cpp` and `sb::Tensor::strides()` (`ptrdiff_t`) as claimed.

---

PR #207, opened 2026-07-02
Handle LHS/RHS aliasing in in-place tensor arithmetic

### LINKED ISSUE #191: [arithmetic] In-place tensor arithmetic ignores LHS/RHS aliasing — a[1:] += a[:-1] silently produces traversal-order-dependent results
## Summary

`include/sbear/tensor.tpp` `broadcast_binop_inplace` (~lines 429-441, used by `operator+=/-=/*=//=`):

```cpp
auto rhs_view = rhs.broadcast_to(out_shape);
sb::walk(lhs, [&](const std::vector<size_t>& idx) {
  lhs(idx) = op(lhs(idx), rhs_view(idx));
});
```

walks `lhs` in row-major order while reading `rhs_view`, with **no overlap check**. If `rhs` is a view of the same buffer traversed differently, later reads observe already-updated elements.

Notably, `assign_from` (tensor.tpp ~147-160) explicitly detects exactly this aliasing case (compares buffer bases, then copies the RHS first) — that was the fix for #6. The in-place arithmetic path was missed, so the same class of corruption survives there:

## Failure scenario

```python
a = sb.tensor(np.array([1., 2., 3., 4.]))
a[1:] += a[:-1]     # NumPy: [1, 3, 5, 7]; here: cascading reads of updated cells → [1, 3, 6, 10]
a += a[::-1]        # order-dependent garbage vs NumPy's [5, 5, 5, 5]
```

Silently wrong values, no error. (NumPy handles both correctly.)

## Suggested fix

Reuse the aliasing detection from `assign_from` in `broadcast_binop_inplace`: same-buffer check → copy `rhs_view` before the walk. Related: #61 asks for a public `shares_memory` helper; this is the internal counterpart.

bwehlin:
`broadcast_binop_inplace` (behind `+=`, `-=`, `*=`, `/=`) walked the LHS while reading an RHS view with no overlap check, so `a[1:] += a[:-1]` or `a += a[::-1]` read elements already updated and silently produced traversal-order-dependent results. It now detects the shared buffer (same `data() - offset()` base comparison as `assign_from`, the #6 fix) and materializes a copy of the RHS first, matching NumPy. The non-aliasing path is unchanged.

New Python regression tests (`test_bugscan_inplace_alias.py`): shifted-overlap add/multiply, reversed-self add, 2-D row overlap, and a non-aliasing control, across float32/64 and int32/64 tensors, plus a PCF-tensor case. Verified to fail against the unfixed code.

Fixes #191

---

### DIFF
--- a/include/sbear/tensor.tpp
+++ b/include/sbear/tensor.tpp
@@ -434,6 +434,21 @@ namespace sb
         throw std::invalid_argument("Cannot broadcast in-place: output shape " +
             shape_to_string(out_shape) + " differs from LHS shape " + shape_to_string(lhs.shape()));
       auto rhs_view = rhs.broadcast_to(out_shape);
+
+      // If the RHS aliases the LHS storage (e.g. a[1:] += a[:-1], where both
+      // sides are views of the same buffer), the walk below would read
+      // elements it has already updated. Materialize an independent copy of
+      // the RHS first, matching NumPy and assign_from; data() includes the
+      // view offset, so compare the buffer base (data() - offset()).
+      if (lhs.data() - lhs.offset() == rhs.data() - rhs.offset())
+      {
+        Tensor<T> rhs_copy = rhs_view.copy();
+        sb::walk(lhs, [&](const std::vector<size_t>& idx) {
+          lhs(idx) = op(lhs(idx), rhs_copy(idx));
+        });
+        return lhs;
+      }
+
       sb::walk(lhs, [&](const std::vector<size_t>& idx) {
         lhs(idx) = op(lhs(idx), rhs_view(idx));
       });
--- a/test/python/test_bugscan_inplace_alias.py
+++ b/test/python/test_bugscan_inplace_alias.py
@@ -1,0 +1,63 @@
+import numpy as np
+import numpy.testing as npt
+import pytest
+
+import stablebear as sb
+from stablebear.base_tensor import FloatTensor, IntTensor
+
+
+# ---------------------------------------------------------------------------
+# Bug #191: in-place tensor arithmetic ignored LHS/RHS aliasing. When the RHS
+# is a view of the same buffer (a[1:] += a[:-1], a += a[::-1]), the in-place
+# walk read elements it had already updated, producing traversal-order-
+# dependent results. NumPy materializes the RHS first; so do we now.
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("np_dtype,TensorT", [(np.float32, FloatTensor),
+                                              (np.float64, FloatTensor),
+                                              (np.int32, IntTensor),
+                                              (np.int64, IntTensor)])
+class TestInplaceAliasing:
+    def test_shifted_overlap_add(self, np_dtype, TensorT):
+        a = TensorT(np.array([1, 2, 3, 4], dtype=np_dtype))
+        a[1:] += a[:-1]
+        expected = np.array([1, 2, 3, 4], dtype=np_dtype)
+        expected[1:] = expected[1:] + expected[:-1].copy()
+        npt.assert_array_equal(np.asarray(a), expected)  # [1, 3, 5, 7]
+
+    def test_reversed_self_add(self, np_dtype, TensorT):
+        a = TensorT(np.array([1, 2, 3, 4], dtype=np_dtype))
+        a += a[::-1]
+        npt.assert_array_equal(np.asarray(a), [5, 5, 5, 5])
+
+    def test_shifted_overlap_mul(self, np_dtype, TensorT):
+        a = TensorT(np.array([1, 2, 3, 4], dtype=np_dtype))
+        a[1:] *= a[:-1]
+        npt.assert_array_equal(np.asarray(a), [1, 2, 6, 12])
+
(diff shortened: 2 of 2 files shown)


include/sbear/tensor.tpp
```diff
@@ -434,6 +434,21 @@ namespace sb
         throw std::invalid_argument("Cannot broadcast in-place: output shape " +
             shape_to_string(out_shape) + " differs from LHS shape " + shape_to_string(lhs.shape()));
       auto rhs_view = rhs.broadcast_to(out_shape);
+
+      // If the RHS aliases the LHS storage (e.g. a[1:] += a[:-1], where both
+      // sides are views of the same buffer), the walk below would read
+      // elements it has already updated. Materialize an independent copy of
+      // the RHS first, matching NumPy and assign_from; data() includes the
+      // view offset, so compare the buffer base (data() - offset()).
+      if (lhs.data() - lhs.offset() == rhs.data() - rhs.offset())
```
bwehlin, line 443: **Minor (residual gap, pre-existing):** this fixes the tensor-RHS overloads, but the element/scalar-RHS in-place overloads just above (`operator+=(const U& u)` etc., ~lines 169-258) still have the same read-after-write hazard when `u` is a reference into the tensor's own buffer, e.g. from C++:

```cpp
t += t(0);  // apply() captures u by reference; after cell 0 is updated,
            // every later cell adds the *doubled* value
```

The first visited element mutates the referent of `u`, so all subsequent elements read the updated value — the same traversal-order-dependent corruption class as #191, just via the scalar path. It is unreachable from Python (`_get_element` returns a copy via pybind's default lvalue-ref policy), so severity is low, but since the PR title claims in-place tensor arithmetic aliasing generally, worth either a `if (&u aliases buffer) copy first` guard or at least a comment on the scalar overloads noting the C++-side caveat.

For the tensor-RHS path itself I could not break the check: `data() - offset()` is exactly `m_data.get()`, every view constructor (`extract`, `broadcast_to`, `reshape`, `squeeze`, `transpose`, `expand_dims`, `flatten`) shares `m_data` with absolute offsets, numpy ingestion (`convert_numpy_to_tensor`) always deep-copies, and nothing constructs an aliasing `shared_ptr` — so base equality is exact (no false negatives), and the copy path is safe even for shallow-copying element types (point clouds, distance/symmetric matrices) because the walk only rebinds cells with freshly constructed `op` results and never mutates payloads in place.

---

test/python/test_bugscan_inplace_alias.py
```diff
@@ -0,0 +1,63 @@
...
+# dependent results. NumPy materializes the RHS first; so do we now.
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("np_dtype,TensorT", [(np.float32, FloatTensor),
+                                              (np.float64, FloatTensor),
+                                              (np.int32, IntTensor),
+                                              (np.int64, IntTensor)])
+class TestInplaceAliasing:
+    def test_shifted_overlap_add(self, np_dtype, TensorT):
```
bwehlin, line 22: **Minor (test coverage):** all overlap patterns exercised here are same-stride shifted or reversed 1-D/row views. A transposed self-overlap is a genuinely different pattern (full-buffer overlap with *different* strides on each side, upper triangle read after lower triangle written) and is the classic case NumPy's overlap detection exists for — it is handled by this fix (same base) but never tested:

```python
def test_transposed_self_add(self, np_dtype, TensorT):
    arr = np.arange(1, 10, dtype=np_dtype).reshape(3, 3)
    a = TensorT(arr.copy())
    a += a.transpose()
    npt.assert_array_equal(np.asarray(a), arr + arr.T)
```

I verified by hand-simulating the pre-fix walk that this fails without the fix (e.g. `a[1][0]` gets `4+2=6`, then `a[0][1]` reads the updated `6` instead of `4`). Also `-=` and `/=` share `broadcast_binop_inplace` so their omission is low-risk, but a `-=` case would be one extra parametrize line and would catch a future refactor that splits the operators. (Note `/=` deliberately raises for `IntTensor`, so only float dtypes could cover it.)

---

bwehlin (commented):
Adversarial review of the aliasing fix. Attack angles tried: (1) defeating the `data() - offset()` base comparison via views-of-views, sliced/transposed/flattened views, reshape of non-contiguous views, and zero-copy numpy ingestion — no false negative found (`data() - offset()` is exactly `m_data.get()`; all view constructors share `m_data` with absolute offsets; `convert_numpy_to_tensor` always deep-copies; nothing builds an aliasing `shared_ptr`); (2) breaking the copy path for broadcast views and heap-payload element types (Pcf, point clouds, distance/symmetric matrices) — safe, since `Pcf` is value-semantic and the walk only rebinds cells with fresh `op` results; (3) checking that the new tests actually fail pre-fix — they do (hand-simulated the unfixed walk for each case); (4) the scalar-RHS path is unaffected by the change as claimed.

Two minor findings posted inline: a residual (pre-existing, C++-only) aliasing gap in the element/scalar-RHS in-place overloads, and a test-coverage suggestion (transposed self-overlap, `-=`). Neither blocks the fix; the core change is correct.

---

PR #208, opened 2026-07-02
Fix axis position for advanced indices following a rank>1 index array

### LINKED ISSUE #182: [indexing] Multiple advanced indices with a >1-D index array gather along the wrong axis (silently wrong results)
## Summary

`_tensor_base.py` `_getitem_entries` (~lines 360-372) applies multiple advanced indices sequentially with outer/np.ix_-style semantics. For each advanced index it computes the target axis as:

```python
dims_dropped = sum(1 for j in range(orig_pos) if isinstance(entries[j], int))
axis = orig_pos - dims_dropped
```

But `_index_select_nd` (~line 427) *inserts* `arr.ndim` output axes at the selected position when the index array has rank > 1 (`new_shape = rshape[:axis] + list(arr.shape) + rshape[axis+1:]`). The loop never accounts for those inserted axes, so every subsequent advanced index is applied at a stale position.

## Failure scenario

For `t` of shape `(3, 4)` wrapping `np.arange(12)`:

```python
t[np.array([[0, 1], [1, 0]]), np.array([1])]
```

- First index (2-D) selects on axis 0 → intermediate shape `(2, 2, 4)`.
- Second index computes `axis = 1` — which is now a **dimension of the first index array**, not the original size-4 axis.
- Result: shape `(2, 1, 4)` with values gathered from the wrong dimension, instead of the expected `(2, 2, 1)` under the library's own ix_-generalized semantics. Silent wrong data; for larger index values it can instead raise a spurious `IndexError` ("index out of bounds for axis 1 with size 2").

The 1-D case is unaffected (`index_select` preserves rank there), which is why existing tests — which only exercise 1-D arrays in outer combinations — pass.

## Suggested fix

Track an axis offset: after applying an advanced index of rank `r` at position `p`, subsequent advanced-index positions greater than `p` must be shifted by `r - 1`. Alternatively, reject rank>1 index arrays in the multi-advanced-index path with a clear error until the semantics are decided.

bwehlin:
`_getitem_entries` applied multiple advanced indices sequentially but never accounted for the output axes a rank-r index array inserts in place of the selected axis, so every subsequent advanced index was applied at a stale position — gathering from a dimension of the first index array instead of the intended axis (silently wrong data, or a spurious `IndexError` for larger index values). The accumulated axis shift (`rank - 1` per applied integer index array) is now tracked; the 1-D case is unaffected.

For `t` of shape `(3, 4)`: `t[np.array([[0, 1], [1, 0]]), np.array([1])]` now returns shape `(2, 2, 1)` with the correct outer-semantics values instead of `(2, 1, 4)` gathered from the wrong dimension.

New regression tests in `test_advanced_indexing.py` (rank-2 then rank-1, two rank-2 arrays, rank-1 then rank-2); the first two were verified to fail against the unfixed code.

Fixes #182

---

### DIFF
--- a/stablebear/_tensor_base.py
+++ b/stablebear/_tensor_base.py
@@ -362,13 +362,17 @@ class Tensor(ABC):
         basic = [slice(None) if isinstance(s, (BoolTensor, IntTensor)) else s
                  for s in entries]
         result = self._basic_getitem(basic, allow_scalar=False)
+        axis_shift = 0
         for orig_pos, idx in advanced:
             dims_dropped = sum(1 for j in range(orig_pos) if isinstance(entries[j], int))
-            axis = orig_pos - dims_dropped
+            axis = orig_pos - dims_dropped + axis_shift
             if isinstance(idx, BoolTensor):
                 result = self._to_py_tensor(result._data.axis_select(axis, idx._data))  # type: ignore[arg-type]
             else:
                 result = self._index_select_nd(result, axis, idx)
+                # A rank-r index array replaces the selected axis with r axes,
+                # so later advanced indices sit r - 1 positions further right.
+                axis_shift += max(idx.ndim, 1) - 1
         return result
 
     def _basic_getitem(self, entries, allow_scalar=True):
--- a/test/python/test_advanced_indexing.py
+++ b/test/python/test_advanced_indexing.py
@@ -210,6 +210,35 @@ class TestMultipleIntIndices:
         arr = np.arange(20, dtype=np.float32).reshape(4, 5)
         _assert_outer_index(arr, (np.array([1, 3]), slice(1, 4)))
 
+    def test_2d_index_array_then_1d(self):
+        # Regression for #182: a rank-2 index array inserts two output axes,
+        # so the next advanced index must shift right by one; it used to be
+        # applied to an axis of the first index array instead.
+        arr = np.arange(12, dtype=np.float32).reshape(3, 4)
+        result = np.asarray(_sb(arr)[np.array([[0, 1], [1, 0]]), np.array([1])])
+        expected = arr[np.array([[0, 1], [1, 0]])][:, :, [1]]
+        assert result.shape == (2, 2, 1)
+        np.testing.assert_array_equal(result, expected)
+
+    def test_two_2d_index_arrays_outer(self):
+        arr = np.arange(12, dtype=np.float32).reshape(3, 4)
+        rows = np.array([[0, 1], [2, 0]])
+        cols = np.array([[1, 3], [0, 2]])
+        result = np.asarray(_sb(arr)[rows, cols])
+        # Outer semantics: each rank-2 index contributes its own two axes.
+        expected = arr[rows][..., cols]
+        assert result.shape == (2, 2, 2, 2)
+        np.testing.assert_array_equal(result, expected)
+
+    def test_1d_then_2d_index_array(self):
+        arr = np.arange(12, dtype=np.float32).reshape(3, 4)
+        rows = np.array([2, 0])
+        cols = np.array([[1, 3], [0, 2]])
+        result = np.asarray(_sb(arr)[rows, cols])
+        expected = arr[rows][:, cols]
+        assert result.shape == (2, 2, 2)
+        np.testing.assert_array_equal(result, expected)
+
 
 class TestMixedBoolIntIndices:
     def test_bool_rows_int_cols(self):


stablebear/_tensor_base.py
```diff
@@ -362,13 +362,17 @@ def _getitem_entries(self, entries, allow_scalar=True):
...
             dims_dropped = sum(1 for j in range(orig_pos) if isinstance(entries[j], int))
-            axis = orig_pos - dims_dropped
+            axis = orig_pos - dims_dropped + axis_shift
             if isinstance(idx, BoolTensor):
                 result = self._to_py_tensor(result._data.axis_select(axis, idx._data))  # type: ignore[arg-type]
             else:
                 result = self._index_select_nd(result, axis, idx)
+                # A rank-r index array replaces the selected axis with r axes,
+                # so later advanced indices sit r - 1 positions further right.
+                axis_shift += max(idx.ndim, 1) - 1
```
bwehlin, line 375: **Medium: the same stale-position bug survives for `newaxis`/scalar-bool inserts after a rank>1 index array.**

`axis_shift` is local to `_getitem_entries`, but `_normalize_index` computes insert positions with `out_pos` incrementing by exactly 1 per `'adv'` entry, and `__getitem__` applies `expand_dims(pos)` after this function returns. So an insert positioned to the right of a rank-r index array is off by `r - 1` — the new axis lands *inside* the block of axes contributed by the index array:

```python
t = FloatTensor(np.arange(12, dtype=np.float32).reshape(3, 4))
t[np.array([[0, 1], [1, 0]]), None]
# actual:   shape (2, 1, 2, 4)  — newaxis splits the index array's own two axes
# expected: shape (2, 2, 1, 4)  — both NumPy's result and the consistent
#                                 outer-semantics result (t[A][:, :, None, :])
```

Same for a scalar-bool insert (`t[A2d, True]`, and `t[A2d, False]` truncates the wrong axis to length 0) and for `t[A2d, None, B]`, where the gather itself is now correct but the inserted axis is misplaced. I verified these shapes by tracing `_normalize_index` -> `_getitem_entries` -> the `expand_dims` loop with the exact code on this branch (1-D index arrays are unaffected, matching NumPy).

This is pre-existing rather than introduced here, but it is the identical root cause the PR description claims to fix ("never accounted for the output axes a rank-r index array inserts"), just in the second consumer of that assumption. Fixing it needs the shift information to reach `__getitem__`'s insert loop — e.g. shift each insert position by `(rank - 1)` for every rank>1 integer-array entry that consumed an output position to its left (or have `_getitem_entries` return the shift).

---

test/python/test_advanced_indexing.py
```diff
@@ -210,6 +210,35 @@ def test_2d_with_slice(self):
...
+        np.testing.assert_array_equal(result, expected)
+
+    def test_1d_then_2d_index_array(self):
+        arr = np.arange(12, dtype=np.float32).reshape(3, 4)
+        rows = np.array([2, 0])
+        cols = np.array([[1, 3], [0, 2]])
+        result = np.asarray(_sb(arr)[rows, cols])
+        expected = arr[rows][:, cols]
+        assert result.shape == (2, 2, 2)
+        np.testing.assert_array_equal(result, expected)
```
bwehlin, line 240: **Low (test coverage): the new tests only exercise adjacent integer-array pairs on a 2-D tensor.**

Cases that the fix also changes (or should protect) but that have no test:

1. **Boolean mask after a rank-2 index array** — `t[A2d, mask]`. The `axis = ... + axis_shift` line fixes the `BoolTensor` branch too (`axis_select` used to be handed a stale axis, previously raising "mask must be 1D with length matching dimension" or selecting from an index-array dimension), but only the `IntTensor` branch is covered by the new tests.
2. **Interleaved basic index** — `t3[A2d, 1:3, B]` and `t3[A2d, 1, B]` on a 3-D tensor. I traced these to the correct result under the fix (`dims_dropped` + `axis_shift` compose correctly), but nothing pins that down; a regression in either term would go unnoticed.
3. **Negative indices in the *second* array** — e.g. `t[A2d, np.array([-1])]`. Resolution happens in `_index_select_nd` against `result.shape[axis]`, which is only correct now because `axis` is correct; a test would lock in that the later array is validated/resolved against the original axis size rather than an index-array dimension.
4. **`newaxis` after a rank-2 array** — `t[A2d, None]` — currently still wrong (see comment in `_tensor_base.py`); worth a test once fixed.

Also note all three new tests build `expected` from composed stablebear-style outer semantics (`arr[rows][..., cols]` etc.) with hardcoded shape asserts — that's fine given the documented ix_-style semantics, but a brief comment that this intentionally differs from NumPy's broadcast result for `arr[rows, cols]` would prevent someone "simplifying" them to direct NumPy comparisons later.

---

stablebear/_tensor_base.py
bwehlin, line 1: **Low (docs sync + read/write asymmetry).**

1. **Docs**: this PR defines user-visible semantics that were previously broken/undefined — each rank-r integer array in a multi-index expression now contributes its own r output axes. `docs/indexing.rst` ("Multiple index arrays") only documents 1-D arrays in outer combinations, and the "multi-dimensional integer index array" paragraph only covers the single-index case. Per the repo's CLAUDE.md doc-sync rule ("user-visible behavior changes should be documented"), a sentence + example for rank>1 arrays in combinations belongs in this PR.

2. **`__setitem__` asymmetry**: the write path (`_setitem_entries`) does *not* have the stale-axis bug — selectors are applied simultaneously against the same basic view via `outer_assign`, so no shift is needed there. But after this PR, `t[A2d, B]` reads correctly while `t[A2d, B] = v` (and even single-index `t[A2d] = v`) is rejected in C++ (`resolve_selectors`: "outer indexing: selector must be 1D"; `validate_axis_indices`: "indices must be 1D"). Pre-existing, and failing loudly is fine — but the `__setitem__` docstring says it "Accepts the same index objects as `__getitem__`", and `docs/indexing.rst` claims "Assignment works with any combination of indices". Both claims are now (more visibly) false for rank>1 arrays; worth either a docs caveat or a clearer Python-level `IndexError`/`NotImplementedError` for rank>1 index arrays in assignment.

---

bwehlin (commented):
Adversarial review of the axis-shift fix. What I attacked and confirmed **holds up**: interleaved slices (`t3[A2d, 1:3, B]`), integers between advanced indices (`t3[A2d, 1, B]` — `dims_dropped` and `axis_shift` compose correctly), negative/out-of-range values in later index arrays (now resolved against the correct original axis), boolean masks following a rank-2 array (the shift fixes the `axis_select` branch too), 1-D behavior (unchanged, NumPy-parity), and `__setitem__` (no equivalent bug — selectors apply simultaneously via `outer_assign`). The outer/ix_-style semantics the fix bakes in match the documented design (`__getitem__` docstring and the indexing guide), so no complaint there. I traced shapes/values through a faithful re-implementation of `_normalize_index`/`_getitem_entries` against NumPy since the extension isn't built here.

What did **not** hold up: the identical stale-position assumption survives in `_normalize_index`/`__getitem__`'s newaxis-insert handling — `t[A2d, None]` inserts the new axis *inside* the index array's contributed axes, shape `(2, 1, 2, 4)` instead of `(2, 2, 1, 4)`. Details plus test-coverage and doc-sync notes in the inline comments (1 medium, 2 low).

---

PR #209, opened 2026-07-02
Fix _gpu_detect import path, DRM double-counting, and WSL2 detection

### LINKED ISSUE #173: [gpu-detect] Compiled _gpu_detect module is dead code — gpu.py imports it from the wrong path (plus DRM double-count and WSL2 gap)
## 1. The C++ detector never runs — wrong import path

`stablebear/gpu.py:8`:

```python
from _gpu_detect import detect_nvidia_gpus, has_nvidia_gpu, nvidia_gpu_count
```

This is a top-level absolute import, but CMake installs the extension *inside* the package — `CMakeLists.txt:1022` (`install(TARGETS ${CPP_MODULE_NAMES} _gpu_detect DESTINATION ${INSTALL_TARGET})` → `site-packages/stablebear/`), which `.github/scripts/compare_wheels.py` also confirms by expecting `stablebear/_gpu_detect*.so`. Since `site-packages/stablebear/` is never on `sys.path`, the import **always** raises `ImportError` and the subprocess fallback (`lspci` / PowerShell / sysfs) silently takes over.

Consequences:
- The purpose-built C++ detection module (`src/gpu_detect/`) is compiled, shipped in every wheel, and never executed.
- In slim Linux containers without `lspci` and without `/sys/class/drm` entries, a real NVIDIA GPU is silently not detected → `_sb_cpp.py` loads the CPU backend with **no warning** (the RuntimeWarning at `_sb_cpp.py:185` fires only when detection succeeds but backend loading fails).
- On Windows, every `import stablebear` pays PowerShell process-spawn latency for detection.

Fix: `from ._gpu_detect import ...` (keeping the existing fallback for source checkouts without the built extension).

Found independently by two review passes; verified against the CMake install destination.

## 2. Linux DRM enumeration double-counts GPUs (latent until #1 is fixed)

`src/gpu_detect/gpu_detect.cpp:33-98` counts every `/sys/class/drm` entry whose `device/vendor` file reads the NVIDIA vendor ID. Each physical GPU exposes both a `cardN` and a `renderDN` node (and connector nodes resolve to the same device), so `nvidia_gpu_count()` reports at least 2× the real count. Backend selection only uses the boolean `has_nvidia_gpu()`, so today the impact is limited to the user-facing count — but worth fixing before enabling the module. Filtering entries to `^card[0-9]+$` fixes it.

## 3. WSL2: both paths come up empty (friction)

On WSL2 the GPU is exposed via `/dev/dxg` (no `/sys/class/drm` PCI entries, no useful `lspci`), yet CUDA works. Both the native path and the fallback report no GPU → silent CPU fallback with no message. Checking `/proc/driver/nvidia/gpus` or `/dev/dxg` existence would cover this common setup.

bwehlin:
- `gpu.py` imported `_gpu_detect` as a top-level module, but CMake installs the extension inside the package (`site-packages/stablebear/`), so the import always failed and the compiled detector shipped in every wheel was dead code — the subprocess fallback silently took over, missing GPUs in slim containers with no `lspci`/drm entries. Now imported relatively (`from ._gpu_detect import ...`), keeping the fallback for source checkouts.
- The Linux DRM scan — in both the C++ module and the Python fallback — counted every `/sys/class/drm` node with an NVIDIA vendor id; each physical GPU exposes `cardN` plus `renderDN`/connector nodes, so counts were at least doubled. Both now count only `cardN` nodes.
- Neither path detected GPUs in slim containers or under WSL2 (`/dev/dxg`, no NVIDIA PCI drm nodes). Both now also enumerate `/proc/driver/nvidia/gpus` (one directory per GPU, with the model name from `information`) and detect WSL2 via `/dev/dxg` plus the WSL NVIDIA driver libraries (the extra check avoids false positives on non-NVIDIA WSL systems).

Fixes #173

## Verification

With this change the compiled module actually loads: `stablebear.gpu.detect_nvidia_gpus` resolves to the C++ extension and correctly reports zero GPUs in the (GPU-less) build container. The new procfs/WSL2 paths are review-verified; no NVIDIA hardware was available to exercise them end-to-end.

---

### DIFF
--- a/src/gpu_detect/gpu_detect.cpp
+++ b/src/gpu_detect/gpu_detect.cpp
@@ -8,6 +8,9 @@
 #include <dirent.h>
 #include <fstream>
 #include <sstream>
+#include <cctype>
+#include <cstring>
+#include <unistd.h>
 #elif defined(_WIN32)
 #include <windows.h>
 #include <dxgi.h>
@@ -30,7 +33,22 @@ struct GpuInfo
 
 #ifdef __linux__
 
-std::vector<GpuInfo> detect_gpus()
+// Each physical GPU exposes several /sys/class/drm nodes (cardN, renderDN,
+// and per-connector nodes that resolve to the same device). Count only the
+// cardN nodes so one GPU is reported once.
+bool is_drm_card_entry(const char* name)
+{
+  if (std::strncmp(name, "card", 4) != 0 || name[4] == '\0')
+    return false;
+  for (const char* p = name + 4; *p != '\0'; ++p)
+  {
+    if (!std::isdigit(static_cast<unsigned char>(*p)))
+      return false;
+  }
+  return true;
+}
+
+std::vector<GpuInfo> detect_gpus_drm()
 {
   std::vector<GpuInfo> gpus;
 
@@ -41,6 +59,9 @@ std::vector<GpuInfo> detect_gpus()
   const struct dirent* entry;
   while ((entry = readdir(drm)) != nullptr)
   {
+    if (!is_drm_card_entry(entry->d_name))
+      continue;
+
     std::string base = "/sys/class/drm/";
     base += entry->d_name;
     std::string device_path = base + "/device/";
@@ -95,6 +116,94 @@ std::vector<GpuInfo> detect_gpus()
 
   closedir(drm);
   return gpus;
+}
+
+// The driver's procfs has one directory per GPU. Covers slim containers
+// with no lspci and no NVIDIA PCI drm entries.
+std::vector<GpuInfo> detect_gpus_proc_driver()
+{
+  std::vector<GpuInfo> gpus;
+
+  DIR* dir = opendir("/proc/driver/nvidia/gpus");
+  if (!dir)
+    return gpus;
+
+  const struct dirent* entry;
+  while ((entry = readdir(dir)) != nullptr)
+  {
+    std::string name = entry->d_name;
+    if (name == "." || name == "..")
+      continue;
+
+    GpuInfo gpu;
+    gpu.vendor_id = NVIDIA_VENDOR_ID;
+    gpu.device_id = 0;
+    gpu.name = "NVIDIA GPU [" + name + "]";
+
+    std::ifstream info_file("/proc/driver/nvidia/gpus/" + name + "/information");
+    std::string line;
+    while (std::getline(info_file, line))
+    {
+      if (line.rfind("Model:", 0) == 0)
+      {
+        auto pos = line.find_first_not_of(" \t", 6);
+        if (pos != std::string::npos)
+        {
+          gpu.name = line.substr(pos);
+        }
+        break;
+      }
+    }
+
+    gpus.push_back(std::move(gpu));
+  }
+
+  closedir(dir);
+  return gpus;
+}
+
+// WSL2 exposes the GPU through /dev/dxg with no NVIDIA PCI drm nodes.
+// /dev/dxg alone could be any vendor, so require the WSL NVIDIA driver
+// libraries as well.
+std::vector<GpuInfo> detect_gpus_wsl()
+{
+  std::vector<GpuInfo> gpus;
+
+  if (access("/dev/dxg", F_OK) != 0)
+    return gpus;
+
+  const char* wsl_markers[] = {
+    "/usr/lib/wsl/lib/libcuda.so.1",
+    "/usr/lib/wsl/lib/libcuda.so",
+    "/usr/lib/wsl/lib/nvidia-smi",
+  };
+  for (const char* marker : wsl_markers)
+  {
+    if (access(marker, F_OK) == 0)
+    {
+      GpuInfo gpu;
+      gpu.vendor_id = NVIDIA_VENDOR_ID;
+      gpu.device_id = 0;
+      gpu.name = "NVIDIA GPU (WSL2)";
+      gpus.push_back(std::move(gpu));
+      break;
+    }
+  }
+
+  return gpus;
+}
+
+std::vector<GpuInfo> detect_gpus()
+{
+  auto gpus = detect_gpus_drm();
+  if (!gpus.empty())
+    return gpus;
+
+  gpus = detect_gpus_proc_driver();
+  if (!gpus.empty())
+    return gpus;
+
+  return detect_gpus_wsl();
 }
 
 #elif defined(_WIN32)
--- a/stablebear/gpu.py
+++ b/stablebear/gpu.py
@@ -5,7 +5,9 @@ falling back to a pure-Python implementation using subprocess.
 """
 
 try:
-    from _gpu_detect import detect_nvidia_gpus, has_nvidia_gpu, nvidia_gpu_count
+    # The extension is installed inside the package (site-packages/stablebear/),
+    # so it must be imported relatively; a top-level import never resolves.
+    from ._gpu_detect import detect_nvidia_gpus, has_nvidia_gpu, nvidia_gpu_count
 except ImportError:
     import os
     import platform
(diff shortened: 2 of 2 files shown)


stablebear/gpu.py
```diff
@@ -5,7 +5,9 @@
 """
 
 try:
-    from _gpu_detect import detect_nvidia_gpus, has_nvidia_gpu, nvidia_gpu_count
+    # The extension is installed inside the package (site-packages/stablebear/),
+    # so it must be imported relatively; a top-level import never resolves.
+    from ._gpu_detect import detect_nvidia_gpus, has_nvidia_gpu, nvidia_gpu_count
```
bwehlin, line 10: **User-visible return-type change once the extension actually loads.** Fixing the import flips which implementation of `detect_nvidia_gpus()` runs for every wheel user, and the two implementations do not return the same thing:

- C++ module: `list[GpuInfo]` — attribute access (`g.name`, `g.vendor_id`, `g.device_id`)
- Python fallback: `list[dict]` — `g["name"]` only, no `vendor_id`/`device_id`

Until now the fallback was the only code path that ever ran, so any downstream code (and the public API surface advertised in `docs/gpu.rst`: `gpu.detect_nvidia_gpus()  # Detailed GPU info`) was effectively defined by the dict shape. After this PR, `gpus[0]["name"]` raises `TypeError` on every properly installed wheel, while continuing to work in source checkouts — the worst kind of environment-dependent divergence. `has_nvidia_gpu()`/`nvidia_gpu_count()` (the only in-repo consumers, via `_sb_cpp.py`) are unaffected.

Suggest normalizing the two paths in this PR (e.g., have the fallback return a small `GpuInfo`-like namedtuple, or have the binding return dicts), and documenting the return shape in `docs/gpu.rst`.

---

src/gpu_detect/gpu_detect.cpp
```diff
@@ -97,6 +118,94 @@ std::vector<GpuInfo> detect_gpus()
...
+    "/usr/lib/wsl/lib/nvidia-smi",
+  };
+  for (const char* marker : wsl_markers)
+  {
+    if (access(marker, F_OK) == 0)
+    {
+      GpuInfo gpu;
+      gpu.vendor_id = NVIDIA_VENDOR_ID;
+      gpu.device_id = 0;
+      gpu.name = "NVIDIA GPU (WSL2)";
```
bwehlin, line 187: **WSL2 path hardcodes a count of 1.** `detect_gpus_wsl()` (and the mirrored Python fallback) pushes exactly one `GpuInfo` regardless of how many GPUs are passed through. On a multi-GPU WSL2 machine `nvidia_gpu_count()` reports 1 — an undercount of the same "user-facing count is wrong" class that item 2 of #173 (DRM double-counting) fixes elsewhere in this PR. Note the `/proc/driver/nvidia/gpus` path does not rescue this: in WSL2 the guest uses `dxgkrnl`, not the `nvidia` kernel module, so that procfs tree doesn't exist and the WSL branch is the one that actually runs.

If enumerating via `/usr/lib/wsl/lib/nvidia-smi -L` is out of scope for the C++ module, at minimum this limitation deserves a comment/docs note, since `nvidia_gpu_count()` is public API (`docs/gpu.rst`, `docs/quickstart.rst`).

---

stablebear/gpu.py
```diff
@@ -52,6 +58,40 @@ def _detect_linux():
...
                         pass
 
+        if gpus:
+            return gpus
+
+        # Fallback: the driver's procfs (one directory per GPU). Covers slim
+        # containers without lspci or drm entries.
+        nvidia_proc = "/proc/driver/nvidia/gpus"
+        if os.path.isdir(nvidia_proc):
+            for entry in sorted(os.listdir(nvidia_proc)):
```
bwehlin, line 68: **Unguarded `os.listdir` can crash `import stablebear`; two sync gaps vs the C++ module.**

1. The per-file reads in this block are carefully wrapped in `try/except OSError`, but `os.listdir(nvidia_proc)` itself is not. `/proc/driver/nvidia/gpus` can exist yet be unreadable (hardened/masked-proc containers), in which case `PermissionError` propagates out of `detect_nvidia_gpus()` — and since `_sb_cpp.py:175` calls `has_nvidia_gpu()` at import time, `import stablebear` itself fails in fallback-mode environments. The C++ counterpart is graceful here (`opendir` failure returns an empty list), so this is a behavioral divergence from the "kept in sync" C++ implementation. (The pre-existing drm `os.listdir` has the same latent gap; this PR adds a second instance on a proc path where permission masking is more common.)

2. Minor divergence in `Model:` parsing: for a present-but-empty `Model:` line, C++ keeps the `"NVIDIA GPU [<pci>]"` placeholder (`find_first_not_of` returns `npos`), while Python sets `name = "".strip()` i.e. an empty string.

---

src/gpu_detect/gpu_detect.cpp
```diff
@@ -30,7 +33,22 @@ struct GpuInfo
 
 #ifdef __linux__
 
-std::vector<GpuInfo> detect_gpus()
+// Each physical GPU exposes several /sys/class/drm nodes (cardN, renderDN,
+// and per-connector nodes that resolve to the same device). Count only the
+// cardN nodes so one GPU is reported once.
+bool is_drm_card_entry(const char* name)
```
bwehlin, line 39: **Collision with open PR #200.** #200 (`claude/repo-review-bugs-friction-lgeypw`, posted yesterday) already fixes items 1–2 of #173 and edits the exact same hunks as this PR: the `gpu.py` import line, the `re.fullmatch(r"card\d+", ...)` filter in the Python fallback, and a cardN filter in this loop — but implemented differently there (inline `std::string`-based check inside `detect_gpus`, no `detect_gpus_drm` rename, no helper). The two filters are semantically equivalent, so there's no correctness divergence, but whichever PR lands second will hit textual merge conflicts in both files and needs a rebase that deliberately picks one implementation. Worth coordinating before merge so the cardN logic doesn't end up duplicated or half-merged.

---

bwehlin (commented):
Adversarial review of the fix for #173.

**Verified as correct** (attacked, held up): the relative import with preserved `ImportError` fallback; the cardN filter in both C++ (`is_drm_card_entry`) and Python (`re.fullmatch(r"card\d+", ...)`) — both accept `card0`/`card10+` and correctly exclude `renderD*` and `card0-DP-1`-style connector nodes, and the two are semantically equivalent; the detection chain is strictly fallback-ordered (drm → procfs → WSL), so `/proc/driver/nvidia/gpus` cannot double-count GPUs already found via DRM; a missing/unreadable per-GPU `information` file degrades gracefully to a placeholder name in both implementations; no new import-time latency for wheel users (the compiled module replaces subprocess probing, and the new `access()`/`opendir()` probes only run when earlier stages find nothing).

**Findings** (inline comments):
1. `detect_nvidia_gpus()` silently changes return type for all wheel users (`list[GpuInfo]` vs the fallback's `list[dict]`) — environment-dependent API divergence (stablebear/gpu.py:10).
2. WSL2 path hardcodes GPU count to 1 on multi-GPU systems, and the procfs path can't compensate there (gpu_detect.cpp:187).
3. New unguarded `os.listdir` on `/proc/driver/nvidia/gpus` in the Python fallback can propagate `PermissionError` into `import stablebear`; C++ side is graceful — a sync divergence. Plus a minor empty-`Model:` parsing divergence (stablebear/gpu.py:68).
4. Direct textual collision with open PR #200, which fixes items 1–2 of #173 in the same hunks with a different cardN-filter implementation (gpu_detect.cpp:39).

Also noting: the PR adds ~150 lines of detection logic with no tests, and its own verification section says the procfs/WSL2 paths are review-verified only. The Python fallback chain (cardN regex, `Model:` parsing, fallback ordering) is cheap to unit-test with monkeypatched paths against a tmp directory tree, and `is_drm_card_entry` could be covered in `sb_test`.

---

PR #210, opened 2026-07-02
Fix low-severity core defects: mean over empty dim, empty PCFs, negative barcode times, rank-0 size

### LINKED ISSUE #196: [core] Low-severity C++ core defects: mean over size-0 dim → silent NaN; empty-PCF invariants unenforced; negative barcode times; rank-0 size()/walk() mismatch
Four small core issues, filed together; happy to split on request.

1. **`algorithms/functional/matrix_reduce.hpp:100-110` — `mean()` over a zero-length dimension silently divides by zero.** Reducing a dim of size 0 yields the default PCF, then `ret /= inDimSize` with `inDimSize == 0` produces all-NaN PCFs with no error. `max_element` (same file, lines ~118-133) was given exactly this guard when #25 was fixed; `mean` was not. Result: silent NaNs instead of the clean "Cannot reduce an empty dimension" error.

2. **`functional/pcf.hpp:37-41` — `Pcf(std::vector&&)` enforces no invariants; empty PCFs reach UB downstream.** The constructor accepts any vector (empty, unsorted, first point not at t=0), while `evaluate` and `iterate_rectangles` rely on all three. Concrete path: a truncated/corrupted file gives `read_vector` length 0 → `read_element<PcfT>` (`io/pcf_io.hpp:15-21`) constructs an empty PCF → the first `combine`/`integrate` dereferences an end iterator (`iterate_rectangles.hpp:57`) instead of producing a clean error. Related: #38 (misleading truncated-file error), #90 (empty-PCF representation).

3. **`persistence/betti_curve.hpp:57-71` and `accumulated_persistence.hpp:73-86` — barcode summaries assume nonnegative event times.** `lastTime`/`lastMidpoint` initialize to 0; a user-constructed barcode with negative births (sublevel-set filtrations) emits a `(0, 0)` breakpoint followed by earlier negative ones — an unsorted PCF that silently corrupts later combine/evaluate operations.

4. **`tensor.tpp:1187-1193` vs `algorithms/walk.hpp` — rank-0 tensors: `size()` returns 0 but `walk()` visits one element (and `get_total_size()` returns 1).** The randomized walk overload reserves `tensor.size()` = 0 RNG slots yet consumes one, so a subsequent draw reuses the same seed slot — breaking the documented stream non-overlap guarantee; progress totals and the `parallelEvalThreshold` comparison also miscount rank-0 tensors. Same root inconsistency behind #35's unfilled 0-d generation.

bwehlin:
The four core defects from #196, kept together as filed:

- **`mean()` over a zero-length dimension** divided by zero and silently returned all-NaN PCFs; it now raises the same clean `std::invalid_argument` (Python `ValueError`) as `max_element`. One pre-existing test asserted the silent behavior as an intentional contrast and was updated to expect the error; the `sb.mean` docstring documents it.
- **Empty PCFs** were accepted by `Pcf(std::vector&&)` but are UB in `evaluate`/`iterate_rectangles` (reachable from a truncated file via io `read_element`); the constructor now throws, and `read_element` additionally validates strictly-increasing breakpoint times so corrupted files fail cleanly.
- **Barcode summaries with negative event times** (sublevel-set filtrations): `barcode_to_betti_curve` and `barcode_to_accumulated_persistence` initialized their cursor at 0 and emitted a `(0, 0)` breakpoint followed by earlier negative ones — an unsorted PCF that corrupted later operations. They now start at the earliest event, producing sorted PCFs whose values on the `[0, inf)` evaluation domain are correct.
- **Rank-0 `Tensor::size()`** returned 0 while `walk()` visits one element and `get_total_size()` returns 1, breaking the randomized walk's RNG-slot reservation (documented stream non-overlap) and progress/threshold counts. It now returns 1 (numpy convention), matching the Python-facing `.size`, which already reported 1. The one C++ test asserting the old value was updated (zero-extent tensors still report 0).

New regression tests in `test/test_bugscan_core.cpp` cover all four.

Fixes #196

---

### DIFF
--- a/include/sbear/io/pcf_io.hpp
+++ b/include/sbear/io/pcf_io.hpp
@@ -17,6 +17,18 @@ namespace sb::io::detail
   {
     using point_type = PcfT::point_type;
     auto pts = read_vector<point_type>(is);
+
+    // Validate the invariants downstream algorithms rely on, so a truncated
+    // or corrupted file yields a clean error instead of UB later. Emptiness
+    // is checked by the Pcf constructor itself.
+    for (size_t i = 1; i < pts.size(); ++i)
+    {
+      if (pts[i].t <= pts[i - 1].t)
+      {
+        throw std::runtime_error("Corrupt PCF data: breakpoint times are not strictly increasing");
+      }
+    }
+
     return PcfT(std::move(pts));
   }
 }
--- a/include/sbear/persistence/accumulated_persistence.hpp
+++ b/include/sbear/persistence/accumulated_persistence.hpp
@@ -5,6 +5,8 @@
 #include "barcode.hpp"
 #include "barcode_summary.hpp"
 
+#include <algorithm>
+
 namespace sb::ph
 {
   /**
@@ -71,7 +73,10 @@ namespace sb::ph
     std::vector<PcfPointT> points;
 
     T cumSum = T{0};
-    T lastMidpoint = T{0};
+    // Start at the earliest midpoint when it is negative (sublevel-set
+    // filtrations), mirroring barcode_to_betti_curve -- starting at 0 would
+    // emit an unsorted PCF.
+    T lastMidpoint = std::min(T{0}, entries.front().midpoint);
 
     for (auto const& entry : entries)
     {
--- a/include/sbear/persistence/betti_curve.hpp
+++ b/include/sbear/persistence/betti_curve.hpp
@@ -55,7 +55,10 @@ namespace sb::ph
     std::vector<PcfPointT> points;
 
     T count = 0;
-    T lastTime = T{0};
+    // Start at the earliest event when it is negative (sublevel-set
+    // filtrations); starting at 0 would emit a (0, 0) breakpoint followed by
+    // earlier times -- an unsorted PCF.
+    T lastTime = std::min(T{0}, events.front().time);
 
     for (auto const& event : events)
     {
--- a/include/sbear/tensor.tpp
+++ b/include/sbear/tensor.tpp
@@ -1186,10 +1186,10 @@ namespace sb
   template <typename T>
   [[nodiscard]] size_t Tensor<T>::size() const noexcept
   {
-    if (m_shape.empty())
-    {
-      return 0_uz;
-    }
+    // A rank-0 tensor holds exactly one element (numpy convention): walk()
+    // visits it and get_total_size() returns 1, so size() must agree --
+    // returning 0 made the randomized walk reserve too few RNG slots and
+    // progress/threshold logic miscount.
     return std::accumulate(m_shape.begin(), m_shape.end(), 1_uz, std::multiplies<size_t>());
   }
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -880,6 +880,7 @@ if (BUILD_TESTER)
       test/test_offset_data_manager.cpp
       test/test_walk_random.cpp
       test/test_subdivide.cpp
+      test/test_bugscan_core.cpp
   )
 
   if (BUILD_CUDA_TESTER)
--- a/include/sbear/algorithms/functional/matrix_reduce.hpp
+++ b/include/sbear/algorithms/functional/matrix_reduce.hpp
@@ -99,6 +99,16 @@ namespace sb
   template <typename PcfT>
   Tensor<PcfT> mean(const Tensor<PcfT>& in, size_t dim, Executor& exec = default_executor())
   {
+    // Same guard as max_element below: reducing a zero-length dimension would
+    // otherwise divide by zero and silently produce all-NaN PCFs.
+    check_reduce_dim(dim, in.shape().size());
+    if (in.shape()[dim] == 0)
+    {
+      std::ostringstream oss;
+      oss << "Cannot reduce an empty dimension: dimension " << dim << " has size 0";
+      throw std::invalid_argument(oss.str());
+    }
+
     auto ret = parallel_tensor_reduce(in, dim, [](const typename PcfT::rectangle_type& rect) {
       return rect.f_value + rect.g_value;
     }, exec);
--- a/include/sbear/functional/pcf.hpp
+++ b/include/sbear/functional/pcf.hpp
@@ -37,7 +37,13 @@ namespace sb
     explicit Pcf(std::vector<TimePoint<Tt, Tv>>&& pts)
       : m_points(std::move(pts))
     {
-
+      // evaluate and iterate_rectangles dereference the first point
+      // unconditionally; an empty PCF is UB there, so reject it here with a
+      // catchable error (e.g. a truncated file reaching io read_element).
+      if (m_points.empty())
+      {
+        throw std::invalid_argument("A PCF requires at least one point");
+      }
     }
 
     /// Converting constructor: convert from a Pcf with different precision.
--- a/stablebear/reductions.py
+++ b/stablebear/reductions.py
@@ -71,6 +71,9 @@ def mean(fs: PcfContainerLike, dim: int = 0):
     ------
     IndexError
         If ``dim`` is out of range for the input tensor's number of dimensions.
+    ValueError
+        If the reduced dimension has size 0 (the mean of zero functions is
+        undefined).
     """
     backend, tensor = _resolve_pcf_inputs(_REDUCTIONS_BACKEND_MAP, fs)
     return _to_tensor(backend.mean(tensor._data, _resolve_dim(dim, tensor.ndim)))
--- a/test/python/test_bugscan_reductions.py
+++ b/test/python/test_bugscan_reductions.py
@@ -94,10 +94,11 @@ def test_max_time_empty_reduced_axis_raises_not_segfault():
     """max_time over a size-0 reduced axis must raise cleanly, never SIGSEGV.
 
     Issue #25 (bug scan): ``max_time`` segfaulted on an empty reduction
-    dimension while ``mean`` handled it. Resolved together with #46 —
-    ``max_element`` now rejects an empty reduction range with ``ValueError``
-    (``max`` has no identity over an empty range). ``mean`` still returns a
-    valid reduced tensor over the same empty axes, so the two stay consistent.
+    dimension. Resolved together with #46 — ``max_element`` now rejects an
+    empty reduction range with ``ValueError`` (``max`` has no identity over
+    an empty range). Issue #196: ``mean`` used to divide by zero on the same
+    input and silently return all-NaN PCFs; it now raises the same
+    ``ValueError``, so the two stay consistent.
     """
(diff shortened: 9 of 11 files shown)


include/sbear/io/pcf_io.hpp
```diff
@@ -17,6 +17,18 @@ namespace sb::io::detail
   {
     using point_type = PcfT::point_type;
     auto pts = read_vector<point_type>(is);
+
+    // Validate the invariants downstream algorithms rely on, so a truncated
+    // or corrupted file yields a clean error instead of UB later. Emptiness
+    // is checked by the Pcf constructor itself.
+    for (size_t i = 1; i < pts.size(); ++i)
+    {
+      if (pts[i].t <= pts[i - 1].t)
```
bwehlin, line 26: **[medium] The `<=` check rejects files that were written from PCFs the public API accepts today — previously-valid saved data now fails to load as "corrupt".**

On main, duplicate breakpoint times are *accepted* by the public constructors:

- `construct_pcf` (`src/python/pypcf_support.hpp`) validates order with `std::is_sorted` and a strict `a.t < b.t` comparator, so equal consecutive times pass ("non-decreasing" is even the documented wording in the error message and in `docs/concepts.rst`). That's exactly the gap PR #155 exists to close — and #155 is still open, not merged.
- `from_serial_content` (`src/python/functional/py_make_from_serial_content.cpp`) performs no time-order validation at all (per #155's description this looseness is deliberate).
- `random_pcfs` (`include/sbear/random.hpp`) sorts uniform draws but never dedupes; with `float32` and large `n_points`, equal draws are a realistic birthday-collision event, not a theoretical one.

`write_element` dumps `pcf.points()` verbatim, so any of these PCFs round-trips to disk today and loads fine. After this change, loading that same file throws `"Corrupt PCF data: breakpoint times are not strictly increasing"` — a backward-compatibility break that mislabels API-accepted data as corruption, and it isn't mentioned in the PR description.

Note duplicate-time PCFs are functional in the core algorithms (`upper_bound` in `evaluate` picks the last duplicate; `iterate_rectangles` just emits zero-width rectangles), so this isn't rejecting data that was already unusable.

Suggestion: validate only what the constructors actually enforce today, i.e. reject *decreasing* times (`pts[i].t < pts[i - 1].t`), and tighten to `<=` only together with #155 (ideally with a format-version bump or at least a changelog/docs note about old files).

---

include/sbear/persistence/betti_curve.hpp
```diff
@@ -55,7 +55,10 @@ namespace sb::ph
     std::vector<PcfPointT> points;
 
     T count = 0;
-    T lastTime = T{0};
+    // Start at the earliest event when it is negative (sublevel-set
+    // filtrations); starting at 0 would emit a (0, 0) breakpoint followed by
+    // earlier times -- an unsorted PCF.
+    T lastTime = std::min(T{0}, events.front().time);
```
bwehlin, line 61: **[medium] The negative-first-breakpoint PCFs this now produces break the CUDA integration kernel — CPU and GPU `pdist` disagree on the exact inputs this PR is fixing.**

The CPU side is fine: `integrate`/`combine` call `iterate_rectangles` with `a = 0`, which uses `max_time_iterator_prior_to` to clamp the start iterator to the last breakpoint `<= 0`. But `cuda_pcf_iterate_rectangles` (`include/sbear/cuda/cuda_pcf_kernel.cuh`) has no such clamp — it starts at `fi = gi = 0` with `t = a = 0`, i.e. it assumes `points[0].t == a == 0`, and its cursor `t = max(fpts[fi].t, gpts[gi].t)` can go *negative*, producing negative-width rectangles.

Concrete divergence with two outputs of this function:

- barcode `{(-2, -1)}` → betti curve `(-2,1),(-1,0)`
- barcode `{(-3, -2.5)}` → betti curve `(-3,1),(-2.5,0)`

Both are identically 0 on `[0, inf)`, so CPU L1 `pdist` correctly gives **0**. Kernel trace (`a=0`): iter 1 advances `gi`, emits rect `(0, -2)` with `|1-1| = 0`; iter 2 advances `fi`, emits rect `(-2, -1)` with `|1-0| = 1` → contributes **+1**; iter 3 emits `(-1, b)` with 0. GPU distance = **1**.

So for sublevel-set users (the stated audience of this fix), any `pdist`/`cdist`/`l2_kernel` that dispatches to CUDA silently returns wrong values while CPU returns right ones — and which backend runs depends on GPU availability and `set_cuda_threshold`. The PR's claim that values "on the `[0, inf)` evaluation domain are correct" holds for `evaluate()` and CPU integration only.

Either the CUDA kernel needs the same clamp-to-`a` start as the CPU `max_time_iterator_prior_to`, or these summaries should not emit breakpoints below 0 (see my comment in `accumulated_persistence.hpp` — clamping the curve at `t = 0` would fix this and the representation-contract issue at once, with identical values on `[0, inf)`).

---

include/sbear/persistence/accumulated_persistence.hpp
```diff
@@ -71,7 +73,10 @@ namespace sb::ph
     std::vector<PcfPointT> points;
 
     T cumSum = T{0};
-    T lastMidpoint = T{0};
+    // Start at the earliest midpoint when it is negative (sublevel-set
+    // filtrations), mirroring barcode_to_betti_curve -- starting at 0 would
+    // emit an unsorted PCF.
+    T lastMidpoint = std::min(T{0}, entries.front().midpoint);
```
bwehlin, line 79: **[low] These outputs now violate the library's own documented PCF contract, and the rest of the stack rejects or silently truncates them.**

`docs/concepts.rst` (line 30) states: "the first time must be ``0`` (PCFs are defined on :math:`[0, \infty)`). Breakpoints given out of order, a first time other than ``0``, or any negative time raise a ``ValueError``" — and `construct_pcf` in `src/python/pypcf_support.hpp` enforces exactly that. Concrete fallout for a betti curve / APF built from a sublevel-set barcode (first breakpoint < 0):

1. **`Pcf.astype()` throws.** `stablebear/functional/pcf.py` implements `astype` as `Pcf(self.to_numpy().astype(np_dtype))`, which goes through `construct_pcf` → `ValueError: The first breakpoint must have time t=0 (got t=-2...)`. Same for any user `to_numpy()` → `sb.Pcf(arr)` round-trip.
2. **CPU arithmetic silently truncates.** `combine` iterates from `a = 0`, so `betti1 + betti2`, `sb.mean` over a tensor of these curves, etc. drop every segment below 0 and return a curve starting at `(0, ...)` — the negative-time information the PR carefully preserves disappears on the first arithmetic op, with no warning. Defensible under `[0, inf)` semantics, but surprising and undocumented.
3. **Docs not updated.** Per the repo's CLAUDE.md doc-sync rule, a user-visible representation change (library-produced PCFs may now carry breakpoints at negative times) should be reflected in `docs/concepts.rst` / the persistence docs; as-is, concepts.rst directly contradicts what `barcode_to_betti_curve`/`barcode_to_accumulated_persistence` now return.

An alternative that avoids all three (plus the CUDA divergence flagged in `betti_curve.hpp`): since the evaluation domain is `[0, inf)` anyway, fold all events at `t < 0` into the initial value and emit the first breakpoint *at* `t = 0` (e.g. `(0, count_alive_at_0)`). That yields a sorted PCF, identical values everywhere the function is defined, and keeps the documented invariant intact.

---

include/sbear/tensor.tpp
```diff
@@ -1186,10 +1186,10 @@ namespace sb
...
   {
-    if (m_shape.empty())
-    {
-      return 0_uz;
-    }
+    // A rank-0 tensor holds exactly one element (numpy convention): walk()
+    // visits it and get_total_size() returns 1, so size() must agree --
+    // returning 0 made the randomized walk reserve too few RNG slots and
+    // progress/threshold logic miscount.
     return std::accumulate(m_shape.begin(), m_shape.end(), 1_uz, std::multiplies<size_t>());
```
bwehlin, line 1193: **[low] This silently changes the on-disk serialization format for rank-0 tensors — files written by older versions no longer load.**

`write_contiguous_tensor` and `read_tensor` (`include/sbear/io/tensor_io.hpp`, lines ~188 and ~233) both loop `data() .. data() + tensor.size()`. Consequences of `size()` going 0 → 1 for rank-0:

- **Old files, new reader:** an old build wrote a rank-0 tensor's header followed by *zero* elements; the new reader computes `sz = 1` and tries to read one element past the payload → EOF error if it's the last object in the stream, or a desynced/garbage parse if other objects follow.
- **New files, old reader:** the extra element is left unread in the stream, desyncing any subsequent reads.

To be fair, the old behavior was itself lossy (a saved rank-0 tensor's single element was never written, so loading returned a default-constructed value), so writing the element is the right direction — but this is a format compatibility break beyond the stated scope ("RNG-slot reservation and progress/threshold counts"), it isn't mentioned in the PR description, and there's no save/load round-trip regression test for rank-0 tensors in `test_bugscan_core.cpp` or the io tests. At minimum it deserves a changelog note; ideally a test, and a decision on whether old rank-0 files should be readable (e.g. tolerate EOF at that point).

---

bwehlin (commented):
Adversarial review of the four fixes. What was attacked and what survived:

- **Empty-PCF constructor throw** — audited every `Pcf(vector&&)` construction site (`combine`/`Accumulator`/`parallel_reduce`, `stable_rank`, `random_pcfs`, `from_serial_content`, `construct_pcf`, betti/APF, io, tests). All internal paths provably produce ≥1 point or are already guarded, so the throw introduces no regression. **No finding.**
- **`mean()` empty-dim guard** — matches `max_element`'s exception type and message exactly, guard ordering (`check_reduce_dim` first) is correct, the docstring is updated, and the rewritten Python test loses no coverage. **No finding.**
- **`read_element` strictly-increasing validation** — **1 finding (medium):** rejects duplicate-time PCFs that the public API accepts today (PR #155 is still open), breaking load of previously-written valid files.
- **Barcode negative-time fix** — the new event-loop logic itself is correct (verified all-negative, mixed, all-nonnegative, and t=0 edge cases by hand, including the C++ test's expected values). **2 findings:** (medium) the resulting negative-first-breakpoint PCFs break the CUDA integrate kernel's `points[0].t == 0` assumption → CPU/GPU `pdist` divergence with a concrete counterexample; (low) they also violate the documented PCF contract (`docs/concepts.rst`), break `Pcf.astype()`/`to_numpy` round-trips, and are silently truncated at 0 by CPU arithmetic — undocumented.
- **Rank-0 `size() == 1`** — audited all C++ `Tensor::size()` callers (walk RNG reservation, progress totals, `flat()` iteration, py bindings). The fix is right for those, but **1 finding (low):** `tensor_io` uses `size()` for the element write/read loops, so the on-disk format for rank-0 tensors changes silently and old files fail to load.

4 findings posted inline. Nothing found against the mean() and empty-PCF fixes.

---

PR #211, opened 2026-07-02
Register a no-op limit_gpus on CPU-only builds; fix buffer-protocol offset

### LINKED ISSUE #197: [bindings] Low-severity binding defects: limit_gpus AttributeError on CPU-only builds; latent double-offset in tensor buffer protocol
1. **`stablebear/system.py:52` — `limit_gpus()` raises AttributeError on CPU-only builds.** The Python wrapper unconditionally calls `cpp.limit_gpus(n)`, but the binding is registered only under `#ifdef BUILD_WITH_CUDA` (`src/python/pymodule.cpp:146-148`). The docstring promises "only has an effect if compiled with GPU support" — i.e. a harmless no-op — but the actual behavior on every CPU-only install (all macOS wheels, `BUILD_WITH_CUDA=0` builds) is `AttributeError: module has no attribute 'limit_gpus'`. Fix: guard with `hasattr`/try-except in `system.py`, or register a no-op binding in the CPU build.

2. **`src/python/py_tensor.hpp:162` — latent double-offset in the buffer protocol.** The buffer view is created from `self.data() + self.offset()`, but `Tensor::data()` *already* includes the offset (`include/sbear/tensor.hpp:180`: `return m_data.get() + m_offset`). Currently unreachable because `extract()` always materializes copies with offset 0 (`tensor.tpp:1494` sets `m_isContiguous = false` for views), but the adjacent comment (tensor.tpp:1489-1493) describes contiguous zero-copy views as a planned optimization — the moment that lands, every numpy view of a sliced tensor reads from `data + 2*offset` (wrong values or OOB). Should be just `self.data()`. Worth fixing now while it's free.

bwehlin:
- `limit_gpus` was only bound under `BUILD_WITH_CUDA`, so the documented "only has an effect if compiled with GPU support" behavior was actually an `AttributeError` on every CPU-only install (all macOS wheels, `BUILD_WITH_CUDA=0` builds). A no-op is now bound instead.
- The buffer protocol created views from `data() + offset()`, but `Tensor::data()` already includes the view offset — a latent double-offset that would read wrong memory the moment the planned contiguous zero-copy views land. Now just `data()`.

Fixes #197

## Tests

New `test/python/test_system_limits.py`: `limit_gpus(1)` must resolve on every build (an `AttributeError` fails the test; a `RuntimeError` from a CUDA build with fewer available GPUs is tolerated), plus a `limit_cpus` round-trip. Verified passing on the CPU-only build. The buffer-offset fix is latent (unreachable until zero-copy views land); existing numpy round-trip tests cover the current behavior.

### DIFF
--- a/test/python/test_system_limits.py
+++ b/test/python/test_system_limits.py
@@ -1,0 +1,22 @@
+import pytest
+
+import stablebear as sb
+
+
+def test_limit_gpus_never_raises_attribute_error():
+    """Regression for #197: on CPU-only builds limit_gpus was not bound at
+    all, so the documented no-op was actually an AttributeError."""
+    try:
+        sb.system.limit_gpus(1)
+    except RuntimeError:
+        # A CUDA build may legitimately complain when fewer GPUs are
+        # available than requested; the point is the call must resolve.
+        pass
+
+
+def test_limit_cpus_roundtrip():
+    """limit_cpus is bound on every build and must accept a positive count."""
+    import os
+
+    sb.system.limit_cpus(2)
+    sb.system.limit_cpus(os.cpu_count() or 1)  # restore parallelism
--- a/src/python/py_tensor.hpp
+++ b/src/python/py_tensor.hpp
@@ -158,8 +158,11 @@ namespace sb_py
           std::transform(self.strides().begin(), self.strides().end(), strides.begin(),
               [](ptrdiff_t v) { return static_cast<pybind11::ssize_t>(v * sizeof(T)); });
 
+          // data() already includes the view offset; adding offset() again
+          // would read past the start of the view (latent until contiguous
+          // zero-copy views land, see tensor.tpp extract()).
           return pybind11::buffer_info(
-              static_cast<void*>(self.data() + self.offset()),
+              static_cast<void*>(self.data()),
               sizeof(T),
               pybind11::format_descriptor<T>::format(),
               self.rank(),
--- a/src/python/pymodule.cpp
+++ b/src/python/pymodule.cpp
@@ -145,6 +145,11 @@ PYBIND11_MODULE(SB_MODULE_NAME, m) {
   m.def("set_min_block_side", [](size_t n){ sb::settings().minBlockSide = n; });
 #ifdef BUILD_WITH_CUDA
   m.def("limit_gpus", [](size_t n){ sb::default_executor().limit_cuda_workers(n); });
+#else
+  // No-op on CPU-only builds, per the Python docstring ("only has an effect
+  // if compiled with GPU support") -- omitting it entirely made
+  // stablebear.system.limit_gpus raise AttributeError instead.
+  m.def("limit_gpus", [](size_t) { });
 #endif
   m.def("get_ngpus", &getNumGpus);
 


test/python/test_system_limits.py
```diff
@@ -0,0 +1,22 @@
+import pytest
+
+import stablebear as sb
+
+
+def test_limit_gpus_never_raises_attribute_error():
+    """Regression for #197: on CPU-only builds limit_gpus was not bound at
+    all, so the documented no-op was actually an AttributeError."""
+    try:
+        sb.system.limit_gpus(1)
```
bwehlin, line 10: **State leak on CUDA builds.** When the build has at least one GPU, `limit_gpus(1)` succeeds and permanently replaces the default executor's CUDA pool with a **1-worker** pool (`Executor::limit_cuda_workers` in `include/sbear/executor.hpp` recreates `m_upCudaExec`; nothing here restores it). Pytest collects this file alphabetically before `test_symmetric_matrix_tensor.py`, `test_tensor_*.py`, etc., so every subsequent GPU-parameterized test in the session (the `device="cuda"` param from `conftest.py`) runs with GPU parallelism capped at 1 on multi-GPU runners — silently changing what the rest of the suite exercises.

The sibling `test_limit_cpus_roundtrip` carefully restores its setting; this test should too, e.g.:

```python
try:
    sb.system.limit_gpus(1)
except RuntimeError:
    pass
else:
    # restore GPU parallelism for the rest of the session
    from stablebear import _sb_cpp as cpp
    sb.system.limit_gpus(cpp.get_ngpus())
```

(restore only on success — if the first call threw, the executor was never replaced, and `limit_gpus(0)` on a CUDA build would itself be problematic).

---

test/python/test_system_limits.py
```diff
@@ -0,0 +1,22 @@
...
+    except RuntimeError:
+        # A CUDA build may legitimately complain when fewer GPUs are
+        # available than requested; the point is the call must resolve.
+        pass
```
bwehlin, line 14: **Over-broad exception tolerance weakens the regression test on CPU builds.** The comment justifies tolerating `RuntimeError` for CUDA builds, but the `except` also swallows it on CPU-only builds — where the documented contract (and the whole point of this PR) is a *harmless no-op*, so any exception there is a regression. If a future CPU binding were changed to `throw std::runtime_error("not supported without CUDA")` (an easy "improvement" someone might make), this test would still pass and #197's user-visible breakage would be back in a different color.

Suggest scoping the tolerance to builds where it can be legitimate:

```python
try:
    sb.system.limit_gpus(1)
except RuntimeError:
    if sb.system.build_type() != "CUDA":
        raise
```

(`build_type()` is already bound on every build.) On CUDA the only throw site is the `nWorkers > get_num_cuda_devices()` check in `executor.hpp::limit_cuda_workers`, so you could optionally also assert on the message to avoid masking unrelated `RuntimeError`s.

---

test/python/test_system_limits.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def test_limit_cpus_roundtrip():
+    """limit_cpus is bound on every build and must accept a positive count."""
+    import os
+
+    sb.system.limit_cpus(2)
+    sb.system.limit_cpus(os.cpu_count() or 1)  # restore parallelism
```
bwehlin, line 22: Nit: "restore parallelism" is approximate rather than a true restore. The default pool is sized with `std::thread::hardware_concurrency()` (`src/executor.cpp:36`), which on recent glibc respects the process CPU affinity mask, while `os.cpu_count()` ignores affinity and reports all logical CPUs. In cpuset/affinity-limited CI containers this line can therefore *raise* the worker count above what the suite started with (mild oversubscription for the remaining tests) instead of restoring it. Harmless today, but worth an accurate comment — or capture the intended count before the first `limit_cpus(2)` call.

---

bwehlin (commented):
Adversarial review of both fixes. What was attacked and verified against the actual sources:

**Buffer-protocol offset fix — verified correct, and genuinely latent as claimed.** `Tensor::data()` does include the view offset (`include/sbear/tensor.hpp:180`, `return m_data.get() + m_offset;`), so the old `data() + offset()` was a double offset. I tried to show it is reachable today and could not: the only code that produces a nonzero `m_offset` is `extract()` (`tensor.tpp:1518/:1591`), which unconditionally sets `m_isContiguous = false` (`tensor.tpp:1494`); every other path (`reshape`, `squeeze`, `flatten`, `copy`) preserves the invariant *contiguous ⟹ offset == 0*, and the `def_buffer` lambda throws on non-contiguous tensors before constructing the `buffer_info`. So pre-change behavior was identical today and the fix is correct future-proofing. No other `data() + offset()` double-offset sites remain in the bindings (checked `py_np_tensor_convert.cpp`, `py_symmetric_matrix.hpp`, `py_barcode.cpp`, `pypcf_support.hpp` — all vector-backed, no tensor offsets).

**No-op `limit_gpus` — verified correct.** `_sb_cpp.py` copies all backend attributes, so `sb.system.limit_gpus` resolves on CPU builds; the docstring's "only has an effect if compiled with GPU support" is now accurate rather than a lie; both variants are unnamed `size_t` lambdas so stubgen signatures match across CPU/CUDA builds; no stale checked-in `.pyi` files exist. The CUDA-side binding is unchanged, and its only throw site (`executor.hpp::limit_cuda_workers`, `std::runtime_error` when n > device count) maps to Python `RuntimeError`, so the new test's tolerance is well-typed on GPU runners. The new test file follows the `test_*.py` convention in `test/python/` and is discovered.

**Findings (all in the new test file, none in the production fixes):**
1. `test_limit_gpus_never_raises_attribute_error` leaks global state on CUDA builds — `limit_gpus(1)` replaces the CUDA pool with a 1-worker pool and never restores it, capping GPU parallelism for the rest of the pytest session.
2. The `except RuntimeError: pass` also swallows exceptions on CPU-only builds, where any exception is a regression of exactly the contract this PR fixes — the tolerance should be scoped to CUDA builds.
3. Nit: `limit_cpus(os.cpu_count() or 1)` does not necessarily restore the original `hardware_concurrency()`-sized pool under CPU-affinity-limited CI.

---

PR #212, opened 2026-07-02
Fix CMake defects: appended CXXFLAGS, FATAL_ERROR mode, stale variables

### LINKED ISSUE #168: [build] CMAKE_CXX_FLAGS is overwritten (not appended), silently discarding user/packager-supplied CXXFLAGS
## Summary

`CMakeLists.txt:994`:

```cmake
if (NOT MSVC)
  set(CMAKE_CXX_FLAGS "-Wall -Wextra -Werror=return-type")
endif ()
```

This *overwrites* `CMAKE_CXX_FLAGS` rather than appending, so any flags the user supplied via the `CXXFLAGS` environment variable or `-DCMAKE_CXX_FLAGS=...` are silently dropped for all top-level targets.

## Failure scenario

- `CXXFLAGS="-march=native" pip install .` — the arch flag is discarded with no warning.
- A distro packager's mandatory hardening flags (`-D_FORTIFY_SOURCE=2`, `-fstack-protector-strong`, ...) are silently removed, producing non-compliant binaries.

## Suggested fix

```cmake
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror=return-type")
```

or better, apply the warning flags via `add_compile_options()` / a per-target `target_compile_options()` so `CMAKE_CXX_FLAGS` is left to the user.

### LINKED ISSUE #170: [build] Latent CMake variable/mode bugs: message(FATAL) typo, VERSION_INFO/PROJECT_TITLE/STUB_FILE reference never-set variables
Four small latent defects in the CMake tree, none currently fatal but each a trap:

1. **`CMakeLists.txt:287`** — `message(FATAL "Could not detect version number from pyproject.toml!")`. `FATAL` is not a valid message mode (should be `FATAL_ERROR`), so it prints "FATAL Could not detect..." as a NOTICE and configuration *continues*, later dying at `project()` with a confusing error about an empty VERSION instead of the intended clear abort.

2. **`CMakeLists.txt:760`** — `target_compile_definitions(${MOD_NAME} PRIVATE VERSION_INFO=${PROJECT_VERSION_FULL})` references `PROJECT_VERSION_FULL`, which is never set — the extraction at line 277 sets `PROJ_VERSION_FULL` (no `ECT`). Every module is compiled with an empty `VERSION_INFO=` define. No source currently uses the macro, so it's harmless today, but any future `#ifdef VERSION_INFO` consumer silently gets an empty string.

3. **`version.cpp.in:11`** — `extern const std::string PROJECT_TITLE = "@PROJECT_TITLE@";` — the CMake variable `PROJECT_TITLE` is never defined anywhere in the tree, so the public `sb::PROJECT_TITLE` (declared in `include/sbear/version.hpp:12`) is always the empty string.

4. **`CMakeLists.txt:1216`** — `message(STATUS "!!! Stub file: ${STUB_FILE}")` prints a variable that is never set; the configure log always shows an empty stub file path.

bwehlin:
Grouped CMake fixes (#168 plus the four latent defects from #170):

- **#168** — warning flags now append to `CMAKE_CXX_FLAGS` instead of overwriting it, so user/packager-supplied `CXXFLAGS` (hardening flags, `-march=...`) are no longer silently discarded.
- **#170.1** — `message(FATAL ...)` is not a valid mode: it printed as a NOTICE and configuration continued to a confusing `project()` error. Now `FATAL_ERROR`.
- **#170.2** — `VERSION_INFO` was compiled from the never-set `PROJECT_VERSION_FULL` (the extraction sets `PROJ_VERSION_FULL`), so the define was always empty.
- **#170.3** — `PROJECT_TITLE`, consumed by `version.cpp.in`, was never set, leaving the public `sb::PROJECT_TITLE` empty; it is now set from `PROJ_NAME`.
- **#170.4** — the configure log printed the never-set `STUB_FILE`; it now reports the meaningful `SKIP_STUBGEN` state.

Fixes #168, fixes #170

Verified by a clean configure + full build + test run on this branch.

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -284,7 +284,7 @@ if (PROJECT_BODY MATCHES [[version[ \t]*=[ \t]*"([^"]+)"]])
 
   set(PROJ_VERSION "${PROJ_VERSION_NUMERIC}")
 else ()
-  message(FATAL "Could not detect version number from pyproject.toml!")
+  message(FATAL_ERROR "Could not detect version number from pyproject.toml!")
 endif ()
 
 project(${PROJ_NAME} VERSION ${PROJ_VERSION_NUMERIC} LANGUAGES CXX ${LANG_CUDA})
@@ -528,6 +528,8 @@ add_subdirectory(3rd/pybind11)
 # Adapted from https://community.gigperformer.com/t/getting-cmake-variables-from-c/17711
 string(TIMESTAMP TODAY "%d/%m/%Y")
 set(PROJECT_BUILD_DATE ${TODAY})
+# Consumed by version.cpp.in (sb::PROJECT_TITLE was always empty before).
+set(PROJECT_TITLE "${PROJ_NAME}")
 configure_file(version.cpp.in ${CMAKE_CURRENT_BINARY_DIR}/version.cpp @ONLY)
 
 SET(SB_LIB_SOURCES_CUDA
@@ -757,7 +759,7 @@ foreach(MOD_NAME IN LISTS CPP_MODULE_NAMES)
   target_include_directories(${MOD_NAME} PRIVATE ${SB_PRIVATE_INCLUDES})
   target_include_directories(${MOD_NAME} PUBLIC ${SB_PUBLIC_INCLUDES})
   target_link_libraries(${MOD_NAME} PRIVATE pybind11::headers)
-  target_compile_definitions(${MOD_NAME} PRIVATE VERSION_INFO=${PROJECT_VERSION_FULL})
+  target_compile_definitions(${MOD_NAME} PRIVATE VERSION_INFO=${PROJ_VERSION_FULL})
   sb_apply_cpu_arch_flags(${MOD_NAME})
 endforeach()
 
@@ -991,7 +993,9 @@ foreach(MOD_NAME IN LISTS CPP_MODULE_NAMES)
 endforeach()
 
 if (NOT MSVC)
-  set(CMAKE_CXX_FLAGS "-Wall -Wextra -Werror=return-type")
+  # Append rather than overwrite so user/packager-supplied CXXFLAGS
+  # (hardening flags, -march=..., ...) are not silently discarded.
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror=return-type")
 endif ()
 
 if (POLICY CMP0177)
@@ -1213,7 +1217,7 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND BUILD_TESTER)
   target_link_libraries(sb_test PRIVATE valgrind_flags)
 endif()
 
-message(STATUS "!!! Stub file:          ${STUB_FILE}")
+message(STATUS "!!! Skip stubgen:       ${SKIP_STUBGEN}")
 message(STATUS "!!! Install target:     ${INSTALL_TARGET}")
 
 message(STATUS "!!! C compiler:         ${CMAKE_C_COMPILER}")


CMakeLists.txt
bwehlin, line 1: **Duplication/conflict with open PR #200.** PR #200 (posted yesterday) already contains fixes for #168 and all four #170 defects in this same file, and both PRs are open. Three of the hunks are character-identical (`FATAL_ERROR`, `PROJ_VERSION_FULL`, the `CMAKE_CXX_FLAGS` append line itself), but the two PRs diverge on details:

- **#170.4**: #200 *deletes* the `!!! Stub file:` message; this PR *replaces* it with a `SKIP_STUBGEN` line → textual conflict.
- **#168**: this PR adds two comment lines above the append that #200 doesn't have → textual conflict on the same hunk.
- **#170.3**: #200 sets `PROJECT_TITLE` at line ~265 (right after `set(PROJ_NAME ...)`); this PR sets it at line ~531 before `configure_file`. A naive merge of both leaves `set(PROJECT_TITLE "${PROJ_NAME}")` **twice** (harmless but sloppy), and no textual conflict flags it.

I verified this with a 3-way merge simulation of the two branches' `CMakeLists.txt`: two conflict blocks (CXXFLAGS hunk, STUB_FILE line) plus the silent PROJECT_TITLE duplication. Whichever PR lands first, the other needs a rebase; consider closing one of the two (or stripping the CMake hunks out of #200, which is otherwise a broader grab-bag).

---

CMakeLists.txt
```diff
@@ -757,7 +759,7 @@ foreach(MOD_NAME IN LISTS CPP_MODULE_NAMES)
   target_include_directories(${MOD_NAME} PRIVATE ${SB_PRIVATE_INCLUDES})
   target_include_directories(${MOD_NAME} PUBLIC ${SB_PUBLIC_INCLUDES})
   target_link_libraries(${MOD_NAME} PRIVATE pybind11::headers)
-  target_compile_definitions(${MOD_NAME} PRIVATE VERSION_INFO=${PROJECT_VERSION_FULL})
+  target_compile_definitions(${MOD_NAME} PRIVATE VERSION_INFO=${PROJ_VERSION_FULL})
```
bwehlin, line 762: This only half-fixes #170.2. Nothing in the tree references the `VERSION_INFO` macro (`git grep VERSION_INFO -- src include stablebear` is empty), and the value is an **unquoted** token: the define is now `VERSION_INFO=0.4.4`, which is not usable as a string. The failure mode #170.2 describes — "any future consumer silently gets an empty string" — just becomes "any future consumer gets an invalid token sequence" unless they add the pybind11-skeleton `MACRO_STRINGIFY` idiom, which this repo doesn't have.

Since `sb::PROJECT_VERSION_FULL` (from `version.cpp.in`) already exposes the full version to C++, the cleaner fixes are either delete this dead define, or make it directly consumable: `VERSION_INFO="${PROJ_VERSION_FULL}"` (escaped quotes so the macro expands to a string literal).

---

CMakeLists.txt
```diff
@@ -284,7 +284,7 @@ if (PROJECT_BODY MATCHES [[version[ \t]*=[ \t]*"([^"]+)"]])
 
   set(PROJ_VERSION "${PROJ_VERSION_NUMERIC}")
 else ()
-  message(FATAL "Could not detect version number from pyproject.toml!")
+  message(FATAL_ERROR "Could not detect version number from pyproject.toml!")
```
bwehlin, line 287: The mode fix is correct, but the "confusing late `project()` failure" scenario from #170.1 is only partially closed. This `FATAL_ERROR` fires only when no `version = "..."` line is found at all. If the line exists but isn't `X.Y.Z`-shaped (e.g. `version = "0.4"` or a typo like `"v0.4.4"`), the `string(REGEX MATCH "^[0-9]+\\.[0-9]+\\.[0-9]+" ...)` at line 280 silently leaves `PROJ_VERSION_NUMERIC` empty and you land in exactly the same confusing `project(... VERSION <empty>)` error the issue complained about. A follow-up guard after line 280 would fully close it:

```cmake
if (NOT PROJ_VERSION_NUMERIC)
  message(FATAL_ERROR "Version '${PROJ_VERSION_FULL}' in pyproject.toml is not of the form MAJOR.MINOR.PATCH")
endif ()
```

---

CMakeLists.txt
```diff
@@ -528,6 +528,8 @@ add_subdirectory(3rd/pybind11)
 # Adapted from https://community.gigperformer.com/t/getting-cmake-variables-from-c/17711
 string(TIMESTAMP TODAY "%d/%m/%Y")
 set(PROJECT_BUILD_DATE ${TODAY})
+# Consumed by version.cpp.in (sb::PROJECT_TITLE was always empty before).
+set(PROJECT_TITLE "${PROJ_NAME}")
```
bwehlin, line 532: Nit: this makes `sb::PROJECT_TITLE` byte-identical to `sb::PROJECT_NAME` (both `"stablebear"` via `PROJ_NAME`), so the public API now exposes two constants with the same value. If a distinct human-readable title was ever intended (which is presumably why a separate `PROJECT_TITLE` variable exists in `version.cpp.in`), consider giving it one (e.g. the pyproject `description`, or `"StableBear"`); otherwise it may be cleaner to drop the constant from `version.cpp.in` / `include/sbear/version.hpp` than to ship a duplicate. Not blocking — empty was worse. (Note #200 makes the identical choice, just at a different line.)

---

bwehlin (commented):
Adversarial review of the CMake fixes. Attack angles tried and **cleared**: the `CMAKE_CXX_FLAGS` append is a non-cache `set()`, so it cannot duplicate flags across reconfigures; the leading space when `CXXFLAGS` is empty is harmless; the `NOT MSVC` guard is correct; pybind11 (line 526) and googletest (line 862) are added via `add_subdirectory` *before* the append at line 998, so third-party code does not inherit `-Wall -Wextra -Werror=return-type`; `.cu` compilation uses `CMAKE_CUDA_FLAGS` and is unaffected; `SKIP_STUBGEN` is unconditionally defined (lines 66-70, forced ON under SKBUILD) so the new log line is accurate; `PROJ_NAME` is set before the new `PROJECT_TITLE` line and `version.cpp.in` uses the matching `@PROJECT_TITLE@`; `PROJ_VERSION_FULL` is set on every path that reaches line 762.

What did stick, in descending severity: (1) the entire CMake portion of this PR duplicates open PR #200, with two guaranteed textual conflicts and one silent `PROJECT_TITLE` duplication if both merge (verified via 3-way merge simulation); (2) the `VERSION_INFO` define is still dead code and still not a usable string; (3) the `FATAL_ERROR` fix leaves the malformed-version path (`version = "0.4"`) falling through to the same confusing `project()` error; (4) nit: `sb::PROJECT_TITLE` is now a duplicate of `sb::PROJECT_NAME`. Details inline.

---

PR #213, opened 2026-07-02
Fix TestPyPI artifact name and pass CODECOV_TOKEN to reusable analysis

### LINKED ISSUE #162: [ci] publish-pypitest.yaml downloads nonexistent artifact `cpu-wheels` — TestPyPI publish is broken
## Summary

`BuildAndPublishTestPyPI` (`.github/workflows/publish-pypitest.yaml`) can never publish: both the `compare` and `publish` jobs download an artifact named `cpu-wheels`, but the `build` job calls `wheels.yaml`, which never uploads an artifact by that name.

## Details

- `publish-pypitest.yaml:42-46` and `:66-70` — `actions/download-artifact@v8` with `name: cpu-wheels`
- `publish-pypitest.yaml:30-32` — `build` job uses `./.github/workflows/wheels.yaml`
- `wheels.yaml` only uploads artifacts named:
  - `wheels-src-<os><suffix>` (line 181)
  - `wheels-final-<os>` (lines 188, 229)
  - `wheels` (merged, line 244)

`cpu-wheels` appears to be a leftover from `wheels-tiny.yaml` (which does produce it, but is not called by this workflow).

## Failure scenario

Any `workflow_dispatch` of BuildAndPublishTestPyPI: `check_version`, `analysis`, and `build` succeed (potentially ~an hour of runner time), then `compare` fails at the download step with "Artifact not found", and `publish` never runs. TestPyPI releases via this workflow are impossible in the current state.

## Suggested fix

Change both download steps to `name: wheels` (the merged artifact from `wheels.yaml`), matching what `publish-pypi.yaml` consumes — or call `wheels-tiny.yaml` if the CPU-only tiny build is the intent for TestPyPI.

### LINKED ISSUE #163: [ci] CODECOV_TOKEN is never passed to the reusable analysis.yaml — coverage uploads run tokenless on every PR/publish pipeline
## Summary

`analysis.yaml` references `secrets.CODECOV_TOKEN` in three Codecov upload steps (lines 169, 177, 453), but its `workflow_call:` block (lines 9–13) declares no `secrets`, and none of its callers pass any:

- `ci.yaml:65-69` — `uses: ./.github/workflows/analysis.yaml` with only `inputs`, no `secrets: inherit`
- `publish-pypi.yaml` and `publish-pypitest.yaml` — same

So whenever `analysis.yaml` runs as a reusable workflow — i.e. every PR CI run and every publish pipeline — `secrets.CODECOV_TOKEN` evaluates to empty. Only a direct manual `workflow_dispatch` of Analysis gets the token.

## Impact

With `fail_ci_if_error: false`, the uploads silently fall back to tokenless (rate-limited) upload or fail quietly. This defeats the flags/carryforward design in `codecov.yml` and makes coverage reporting on PRs unreliable, with no visible CI failure to point at the cause.

## Suggested fix

Either add `secrets: inherit` at each call site, or declare the secret explicitly in the reusable workflow:

```yaml
workflow_call:
  inputs: ...
  secrets:
    CODECOV_TOKEN:
      required: false
```

and pass it from the callers.

bwehlin:
Grouped CI-workflow fixes:

- **#162** — `publish-pypitest.yaml`'s `compare` and `publish` jobs downloaded the nonexistent artifact `cpu-wheels` (a leftover from `wheels-tiny.yaml`), so a TestPyPI publish could never get past the build (~an hour of runner time, then "Artifact not found"). Both steps now download the merged `wheels` artifact that `wheels.yaml` uploads, matching what `publish-pypi.yaml` consumes.
- **#163** — `analysis.yaml` references `secrets.CODECOV_TOKEN` in three upload steps, but its `workflow_call` declared no secrets and no caller passed any, so every PR/publish pipeline uploaded coverage tokenless (rate-limited, silently unreliable). The secret is now declared (`required: false`) and passed explicitly from `ci.yaml`, `publish-pypi.yaml`, and `publish-pypitest.yaml`.

Fixes #162, fixes #163

YAML parses cleanly; the fixes will fully prove out on the next CI/publish run.

---

### DIFF
--- a/.github/workflows/analysis.yaml
+++ b/.github/workflows/analysis.yaml
@@ -11,6 +11,9 @@ on:
       use_github_runners:
         type: boolean
         default: false
+    secrets:
+      CODECOV_TOKEN:
+        required: false
 permissions:
   contents: write
 
--- a/.github/workflows/publish-pypitest.yaml
+++ b/.github/workflows/publish-pypitest.yaml
@@ -26,6 +26,8 @@ jobs:
       contents: write
     uses: ./.github/workflows/analysis.yaml
     needs: check_version
+    secrets:
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   build:
     uses: ./.github/workflows/wheels.yaml
@@ -42,7 +44,7 @@ jobs:
       - name: Download wheels
         uses: actions/download-artifact@v8
         with:
-          name: cpu-wheels
+          name: wheels
           path: dist/
 
       - name: Compare wheels
@@ -66,7 +68,7 @@ jobs:
       - name: Download wheels
         uses: actions/download-artifact@v8
         with:
-          name: cpu-wheels
+          name: wheels
           path: dist/
 
       - name: Publish to TestPyPI
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -67,3 +67,5 @@ jobs:
     uses: ./.github/workflows/analysis.yaml
     with:
       use_github_runners: true
+    secrets:
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/publish-pypi.yaml
+++ b/.github/workflows/publish-pypi.yaml
@@ -66,6 +66,8 @@ jobs:
       contents: write
     uses: ./.github/workflows/analysis.yaml
     needs: check_version
+    secrets:
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   publish:
     runs-on: ubuntu-latest


.github/workflows/analysis.yaml
```diff
@@ -11,6 +11,9 @@ on:
       use_github_runners:
         type: boolean
         default: false
+    secrets:
+      CODECOV_TOKEN:
+        required: false
```
bwehlin, line 16: **Missed caller: `analysis-github.yaml` still passes no secret.** This PR fixes 3 of the 4 call sites of `analysis.yaml`, but `.github/workflows/analysis-github.yaml:9` (`Analysis (GitHub runners)`, `workflow_dispatch`) still calls `uses: ./.github/workflows/analysis.yaml` with no `secrets:` block. Because the secret is declared `required: false`, that caller doesn't break — it silently keeps the exact #163 failure mode: its three Codecov uploads run tokenless (rate-limited/unreliable). So #163 is not fully fixed by this PR; a manual dispatch of that workflow still uploads without a token.

For comparison, open PR #200 (which also claims to fix #163) does cover this call site via `secrets: inherit`. Suggest adding here either:

```yaml
    secrets:
      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```

to `analysis-github.yaml`, or `secrets: inherit`.

---

.github/workflows/publish-pypitest.yaml
```diff
@@ -42,7 +44,7 @@ jobs:
       - name: Download wheels
         uses: actions/download-artifact@v8
         with:
-          name: cpu-wheels
+          name: wheels
```
bwehlin, line 47: **Minor / description accuracy:** `name: wheels` is the correct fix — it matches the merged artifact `wheels.yaml` uploads via `actions/upload-artifact/merge@v4` (flat layout, so `dist/*.whl` works for both `compare_wheels.py` and the pypa publish action). However, the PR description's claim that this now "match[es] what `publish-pypi.yaml` consumes" is not quite right: `publish-pypi.yaml` downloads the `dist` artifact (a further merge of `{wheels,sdist}`), not `wheels`. Consequence worth being aware of: a TestPyPI publish exercises wheels only and never uploads/tests an sdist, unlike the real PyPI pipeline — fine if intentional (this workflow has no `sdist` job), but it means TestPyPI runs don't rehearse the full PyPI publish.

---

bwehlin (commented):
Adversarial review of the #162/#163 fixes.

**What was attacked and held up:** all four changed YAML files parse cleanly; `wheels` is the correct merged artifact name (`upload-artifact/merge@v4`, `pattern: wheels-final-*`, flat layout) and the `dist/` download path satisfies both `compare_wheels.py` and `pypa/gh-action-pypi-publish`; the `compare`/`publish` jobs correctly wait on the whole reusable `wheels.yaml` run including `merge_wheel_artifacts`; `required: false` means callers that don't pass the secret degrade instead of fail, so fork-PR/dependabot contexts (empty secret) are safe.

**Findings (2):**
1. **Medium — #163 is only partially fixed.** `analysis-github.yaml` is a fourth caller of `analysis.yaml` and still passes no `CODECOV_TOKEN`; its manual dispatch keeps uploading coverage tokenless (silently, since the secret is optional). See inline comment.
2. **Minor — description inaccuracy** about `publish-pypi.yaml` consuming the same artifact (it consumes the merged `dist`, not `wheels`); the fix itself is correct. See inline comment.

**Overlap warning:** open PR #200 (posted yesterday) also fixes #162 and #163, using `secrets: inherit` at all four call sites (including `analysis-github.yaml`) plus the identical `cpu-wheels` → `wheels` rename. If both are merged, the `secrets:` hunks in `ci.yaml` / `publish-pypi.yaml` / `publish-pypitest.yaml` will conflict. Worth deciding which PR carries these fixes.

---

PR #214, opened 2026-07-03
Added type checks in plot_barcode.

### LINKED ISSUE #159: plot_barcode silently accepts a PointCloud (no input-type validation)
### Summary

`plot_barcode` (`stablebear/plotting.py`) performs no validation that its input is actually a `Barcode`/`BarcodeTensor`. A `PointCloud` (or `PointCloudTensor`) is silently accepted and rendered as a barcode, instead of being rejected with a clear error.

### Where it happens

`plot_barcode(bc, ...)` only special-cases `BarcodeTensor`:

```python
if isinstance(bc, BarcodeTensor):
    ...
    return ...

bars = np.asarray(bc)          # <-- fall-through: any array-like is accepted
if len(bars) == 0:
    return y_offset
...
lengths = np.where(np.isfinite(bars[:, 1]), bars[:, 1] - bars[:, 0], np.inf)
order = np.lexsort((lengths, bars[:, 0]))[::-1]
```

Anything that isn't a `BarcodeTensor` falls through to `np.asarray(bc)` and then has columns `0` and `1` interpreted as `(birth, death)` pairs. A point cloud's per-cloud array has shape `(n_points, dim)`, so with `dim >= 2` it coerces cleanly to an `(n, 2)`-style array and the point coordinates are drawn as bogus `(birth, death)` bars. No `TypeError` is raised — the user gets a meaningless plot rather than an error.

### Expected behavior

`plot_barcode` should validate its input and reject anything that is not a `Barcode` or `BarcodeTensor`, raising a clear `TypeError` (e.g. `"plot_barcode expects a Barcode or BarcodeTensor, got PointCloudTensor"`). A `PointCloud` is not a barcode and should not be plotted as one.

### Suggested fix

Add an explicit type guard before the `np.asarray(bc)` fall-through: accept only `Barcode` / `BarcodeTensor` (and, if intended, raw `(n, 2)` numpy arrays), and raise `TypeError` for everything else — in particular `PointCloud`/`PointCloudTensor`.

### Notes

- File: `stablebear/plotting.py`, `def plot_barcode` (the branch after the `isinstance(bc, BarcodeTensor)` check).
- A regression test asserting `plot_barcode(point_cloud)` raises `TypeError` would guard against this.

MattiasBjerklov:
## Fix
**Python (`stablebear/plotting.py`)**

`plot_barcode` now accepts an `(n, 2) numpy.ndarray` of `(birth, death)` pairs directly, in addition to Barcode and 1-D BarcodeTensor. This mirrors the Barcode constructor, which already treats an (n, 2) array as a barcode.

Input validation added:

- `TypeError` if the input is not a `Barcode`, `BarcodeTensor`, or `numpy.ndarray`, or if an array's dtype is neither floating nor integer.

- `ValueError` if an array is not of shape `(n, 2)`. 

## Tests
Python `stablebear/test/python/test_plotting.py`. Added four tests checking the error messages for unsupported types/shapes.

Closes #159 .

### DIFF
--- a/stablebear/plotting.py
+++ b/stablebear/plotting.py
@@ -70,8 +70,9 @@ def plot_barcode(bc, ax=None, y_offset=0, **kwargs):
 
     Parameters
     ----------
-    bc : Barcode or BarcodeTensor
-        A single ``Barcode`` or a 1-D ``BarcodeTensor``. For a tensor,
+    bc : Barcode, BarcodeTensor or NumPy array
+        A single ``Barcode``, a 1-D ``BarcodeTensor`` or an (n, 2) NumPy array
+        of ``(birth, death)`` pairs. For a tensor,
         the barcodes are stacked vertically in order.
     ax : matplotlib axes, optional
         Axes to plot on. If ``None``, uses ``matplotlib.pyplot`` directly.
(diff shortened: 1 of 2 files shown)


MattiasBjerklov:
I saw that an empty plot is now generated when an unsupported data type is plugged into plot_barcode. I will now fix it so that the plot is generated only for supported data types.

MattiasBjerklov:
It should be fixed now.

PR #216, opened 2026-07-07
Relative subsampling

FilipNorth:
### Things missing:
- Custom filter functions
- Custom distributions
- Optimization for replace=False sampling. Will probably use Fenwick tree but for future feature
- Progress bar and cancellation for initial weight matrix calculation, right now only for the actual sampling. GIL is unlocked.
- Chunked weight evaluation: the dense (n_query × n_reference) weight matrix is O(n²) memory with query=None; docs recommend a query subset for large clouds

### Things which could be improved
- Most tests are AI written
- Docs and doc strings are almost completely AI written
- The plots in the documentation have bad colours (Remembered now, easy fix but will submit now anyway)

### Things which work well
- The sampling is mostly very easy to use
- Sampling weights were restructured to be prepared once per query point and shared across instances (~100× faster draws with replacement at large sizes)
- The sampling is well tested with many validations and works across extreme parameter regimes, for example Gaussian using the Log-Sum-Exp Trick to avoid numerical errors when sampling an extreme number of standard deviations away, so we always have expected behaviour.
   - The Log-Sum-Exp Trick works for every exponential-family distribution we add, not just Gaussian, and has a quite nice solution in my opinion

## Below is a Claude-written PR review guide which is more detailed

# PR draft: Relative subsampling of point clouds and distance matrices

> Draft for the single feature PR from `relative-subsampling` into `main`.
> Assembly: your "Relative subsampling MVP implemented" summary goes on top,
> the body below follows it under its "detailed review guide" heading. Delete
> this file when the PR is open.

**Suggested title:** `Add relative subsampling (stablebear.sampling)`

---

Implements the front end of the *relative approach* of Agerberg, Chachólski &
Ramanujam (2023): for each query point, draw `n_instances` subsamples of up to
`sample_size` points from a reference point cloud (or a precomputed
`DistanceMatrix`), with per-query sampling probabilities formed by passing the
Euclidean (or stored) distance through a built-in distribution — `Gaussian` or
`Uniform`. Weighting and drawing run fused in parallel C++; the result is a
`(n_query, n_instances)` `PointCloudTensor` / `DistanceMatrixTensor` that feeds
directly into `compute_persistent_homology` → `barcode_to_stable_rank` → `mean`
for a relative stable rank per query point.

```python
import stablebear as sb

subs = sb.subsample_relative(reference, query, sample_size=30, n_instances=2000,
                             distribution=sb.Gaussian(sigma=0.5))
bcs = sb.persistence.compute_persistent_homology(subs, max_dim=1)
```

New user guide: `docs/sampling.rst` (background, distributions, figures,
pipeline example); API reference: `docs/api_sampling.rst`.

## Review guide

The PR is large but layered; most of the line count is bindings-per-precision,
tests, and docs. Suggested reading order:

1. **The indexed-view concept** (the one idea everything else leans on):
   `include/sbear/point_cloud.hpp`, `include/sbear/distance_matrix.hpp`.
   A subsample stores only its drawn indices and shares the reference's buffer;
   `n_points()/size()/operator()(i, j)` read through the indexing
   transparently, so Ripser consumes subsamples with no special case.
   `materialize()` produces an owning copy; views are read-only.
2. **Draw primitives**: `include/sbear/sampling/weighted_draw.hpp` — pure
   functions of a weight/CDF row and an engine. A row is prepared once per
   query point (validation + eligible count + in-place prefix sums with
   replacement), then drawn from once per subsample instance.
3. **Weighting**: `include/sbear/sampling/weighting.hpp` — filter and
   distribution functors, the dense `(n_query, n_reference)` weight matrix,
   and the log-space max-shift channel (`log_weight`) that keeps Gaussian rows
   from underflowing to artificial empty regions at large distances.
4. **Orchestration**: `include/sbear/sampling/subsample_task.hpp`,
   `subsample.hpp` — synchronous weights + row preparation, then one
   stoppable, progress-reporting parallel walk over the output grid (the
   codebase's standard single-phase task shape). Each cell derives its RNG
   engine from its flat index in a seed block reserved *synchronously* from
   the caller's generator, so results are independent of thread count and
   consecutive calls draw fresh samples.
5. **Serialization**: `include/sbear/io/tensor_io.hpp`, `io.hpp`,
   `compressed_matrix_io.hpp` — shared-source formats (point clouds
   1000→1001, distance matrices 1120→1121): each distinct source buffer is
   stored once, then per-element (source id, optional index array). Legacy
   files remain readable. A *standalone* indexed `DistanceMatrix` refuses to
   serialize (materialize first); `data()` on a view throws, `source_data()`
   is the deliberate source accessor.
6. **Bindings**: `src/python/sampling/` — descriptor classes + `std::variant`
   dispatch, one instantiation per (filter, distribution, precision). The GIL
   is released around the blocking phases. Mechanical: the 32/64 registration.
7. **Python front end**: `stablebear/sampling/` (input resolution, the
   `Gaussian`/`Uniform` specs), `stablebear/base_tensor.py` (the new
   `PointCloud` element wrapper).

## API changes to note

- **`PointCloudTensor[i, j]` now returns a `PointCloud` wrapper** (was a
  `FloatTensor`). It exposes `.shape`, `.is_indexed`, `.indices`,
  `.materialize()`, `__array__`, indexing, and in-place coordinate writes for
  owning clouds; indexed views are read-only. Code that called FloatTensor
  methods on an element must go through `.materialize()`.
- `DistanceMatrix` gains `.is_indexed`, `.indices`, `.materialize()`.
- New top-level exports: `subsample_relative`, `Gaussian`, `Uniform`,
  `PointCloud`.

## Semantics pinned by tests

- `sample_size` is a **maximum**: without replacement a draw shrinks to the
  eligible count; a query with no positively-weighted reference points yields
  empty subsamples (a `UserWarning` names them when `verbose=True`). Only
  `Uniform` can produce an empty region; `Gaussian` weights every point
  positively at any distance (log-space evaluation).
- Zero-weight points are never drawn, including on the
  `uniform() == cdf.back()` boundary (LWG 2524) and for trailing zeros created
  mid-draw by without-replacement zeroing.
- Consecutive calls advance the caller's (or global) generator; reseeding
  replays the sequence call by call.
- Degenerate distribution parameters (`high <= low`, `sigma <= 0`, non-finite
  `mean`/`sigma`) are rejected at construction, so an empty subsample always
  means a valid region that happens to contain no reference points.
- `allclose` on indexed distance-matrix views compares the principal
  submatrix (like `operator==`), not the raw shared buffer.
- Draws are independent of thread count (per-cell engines seeded from flat
  output indices).

## Rationale behind the "Things missing" list above

- **Custom filters/distributions are deliberately excluded, not just
  unimplemented.** Each built-in owns a fully fused C++ fast path (distances +
  weighting + draw); a Python callable per point pair would put the
  interpreter inside the O(n_query × n_reference) hot loop. Custom
  distributions existed mid-development and were removed to keep the feature
  simple; the `log_weight` channel in `weighting.hpp` is the extension point
  when they return.
- **Dense weight matrix is O(n_query × n_reference).** With `query=None` that
  is n² doubles (50k reference points ≈ 20 GB). Documented; recommended
  mitigation is a query subset. Chunked weight evaluation is the real fix,
  deferred until needed.
- **The weighting/preparation phases are silent and non-cancellable.** They
  run synchronously before the stoppable draw task (the draw phase — the long
  one — has progress + Ctrl-C). The GIL *is* released, so other Python threads
  keep running. Folding precalc into the task would reintroduce the
  multi-phase task machinery deliberately removed during development; that
  remains the escape hatch if the silent window becomes a real UX problem.
- **Without-replacement draws rebuild the CDF per draw** — O(sample_size ×
  n_reference) per instance. Acceptable for a first version; a Fenwick-tree
  rewrite is planned as a follow-up PR, and a seed-pinned regression test
  will be added together with it (bit-exact draws are the gate for that
  refactor).
- **`PointCloud` publicly inherits `Tensor<T>` and adds `m_indices`** —
  copying through the base type slices the indices off. All in-tree call
  sites route through `copy()`/`materialize()`; composition can come later if
  it proves fragile.

## Testing

- C++ (`test/test_weighted_draw.cpp`, `test_tensor.cpp`, `test_tensor_io_core.cpp`):
  row-preparation contracts (eligibility, validation, NaN), draw-length
  contract (0 / min(s, eligible) / s), zero-weight exclusion incl. the CDF
  boundary, worker-exception propagation, `data()` view guard, format ids.
- Python (`test/python/sampling/`, `test_io_roundtrip.py`): end-to-end
  region semantics for both distributions, ragged/empty draws, generator
  advancement + reproducibility, verbose/quiet equivalence, indexed-view
  read-only behavior, distmat parity, save/load round-trips (sharing
  restored, empty index arrays, standalone-view error), full
  subsample → persistence → stable-rank pipeline.
- Suites: 460 C++ / 1892 Python, all green. Docs build clean.


include/sbear/io/compressed_matrix_io.hpp
```diff
@@ -5,12 +5,25 @@
...
   {
+    // An indexed view's size() (selected points) disagrees with its
+    // storage_count()/data() (the full shared source buffer): writing it here
+    // would desynchronize the stream. Views inside tensors are handled by the
+    // shared-source format in tensor_io.hpp; a standalone view must be
+    // materialized by the caller.
+    if constexpr (requires { mat.is_indexed(); })
+    {
+      if (mat.is_indexed())
+        throw std::invalid_argument(
```
bwehlin: Let's just materialize internally and write the result. This keeps things easy to use.

include/sbear/io/tensor_io.hpp
```diff
@@ -31,6 +35,24 @@ namespace sb::io::detail
...
+  struct is_point_cloud : std::false_type {};
+
+  template <typename T>
+  struct is_point_cloud<PointCloud<T>> : std::true_type { using scalar_type = T; };
+
+  template <typename T>
+  inline constexpr bool is_point_cloud_v = is_point_cloud<T>::value;
+
+  template <typename T>
+  struct is_dist_matrix : std::false_type {};
```
bwehlin:
```suggestion
  struct is_distance_matrix : std::false_type {};
```

include/sbear/io/tensor_io.hpp
```diff
@@ -114,14 +136,21 @@ namespace sb::io::detail
     else if constexpr (std::is_same_v<T, Pcf<int32_t, int32_t>>) { return TensorFormat{ .baseFormat = 101, .subFormat = 32 }; }
     else if constexpr (std::is_same_v<T, Pcf<int64_t, int64_t>>) { return TensorFormat{ .baseFormat = 101, .subFormat = 64 }; }
 
-    else if constexpr (std::is_same_v<T, PointCloud<float32_t>>) { return TensorFormat{ .baseFormat = 1000, .subFormat = 32 }; }
-    else if constexpr (std::is_same_v<T, PointCloud<float64_t>>) { return TensorFormat{ .baseFormat = 1000, .subFormat = 64 }; }
```
bwehlin, line 118: We need to add integration tests for the old formats. I.e., using an older stablebear, we write some test point clouds, then in the test suite, we add read tests for these files. This is to ensure that we are not breaking existing flows.

Same thing with DistanceMatrix below.

include/sbear/io/tensor_io.hpp
```diff
@@ -170,6 +199,89 @@ namespace sb::io::detail
...
   }
 
+  // Serialize the elements of a tensor of point clouds by storing each distinct
+  // source coordinate buffer once (point clouds that share a source — e.g. the
+  // indexed subsamples from stablebear.sampling — are deduplicated), followed by
+  // every element as a source id plus, for indexed views, its index array.
+  template <typename ScalarT>
+  void write_point_cloud_elements(std::ostream& os, const Tensor<PointCloud<ScalarT>>& tensor)
+  {
+    std::unordered_map<const void*, uint64_t> idOf;
```
bwehlin: No reason to use void* here, we know what the type is.

include/sbear/io/tensor_io.hpp
```diff
@@ -170,6 +199,89 @@ namespace sb::io::detail
...
+    std::vector<const PointCloud<ScalarT>*> sources;
```
bwehlin: This only deduplicates within a single PointCloudTensor. What if we have several draws made from the same source PointCloud? We need to keep track somehow at a file level rather than an individual PointCloudTensor level.

include/sbear/io/tensor_io.hpp
```diff
@@ -170,6 +199,89 @@ namespace sb::io::detail
...
+
+    write_bytes<uint64_t>(os, static_cast<uint64_t>(sources.size()));
+    for (const PointCloud<ScalarT>* src : sources)
+      write_tensor(os, static_cast<const Tensor<ScalarT>&>(*src));
+
+    for (auto k = 0_uz; k < sz; ++k)
+    {
+      write_bytes<uint64_t>(os, sourceId[k]);
+      const PointCloud<ScalarT>& elem = data[k];
+      write_bytes<uint8_t>(os, elem.is_indexed() ? uint8_t{1} : uint8_t{0});
```
bwehlin: We should do `write_bytes<bool>`. However, we'll have to add a specialization for it since a bool can technically be any value 0-255.

include/sbear/io/tensor_io.hpp
```diff
@@ -170,6 +199,89 @@ namespace sb::io::detail
...
+  // As write_point_cloud_elements, for tensors of distance matrices: each
+  // distinct source buffer is stored once as a full compressed matrix
+  // (uint64 size + entries, the read_compressed_matrix layout), then every
+  // element as a source id plus, for indexed views, its index array. This is
+  // what lets subsampled sub-matrices be saved without either duplicating the
+  // source per element or desynchronizing on size()/storage_count().
+  template <typename ScalarT>
+  void write_distance_matrix_elements(std::ostream& os, const Tensor<DistanceMatrix<ScalarT>>& tensor)
+  {
+    std::unordered_map<const void*, uint64_t> idOf;
```
bwehlin: Can we use a helper shim that deals with both DistanceMatrix and PointCloud?

include/sbear/io/tensor_io.hpp
```diff
@@ -243,6 +366,88 @@ namespace sb::io::detail
...
+    return ret;
+  }
+
+  // Read the current (baseFormat 1121) distance-matrix tensor format: distinct
+  // source matrices stored once, then per-element (source id, optional indices).
+  // Elements referencing the same source share its buffer, as before saving.
+  template <typename ScalarT>
+  Tensor<DistanceMatrix<ScalarT>> read_indexed_distance_matrix_tensor(std::istream& is)
+  {
+    auto shapeSz = read_bytes<std::uint64_t>(is);
```
bwehlin: Same comment about unifying DistanceMatrix and PointCloud as in the writer part.

include/sbear/sampling/subsample.hpp
```diff
@@ -0,0 +1,107 @@
...
+        throw std::invalid_argument("reference must be a 2-D (n_points, dim) point cloud");
+      if (sampleSize == 0)
+        throw std::invalid_argument("sample_size must be positive");
+    }
+
+    template <typename T>
+    void validate_distmat(const DistanceMatrix<T>& source, size_t sampleSize)
+    {
+      if (source.size() == 0)
+        throw std::invalid_argument("reference distance matrix must be non-empty");
```
bwehlin:
```suggestion
        throw std::invalid_argument("reference distance matrix must be nonempty");
```

include/sbear/sampling/subsample.hpp
```diff
@@ -0,0 +1,107 @@
...
+  } // namespace detail
+
+  /// Subsample @p reference relative to each point of @p query, weighted by
+  /// @p distribution of @p filter per point pair. Returns a SubsampleHandle:
+  /// the async draw task plus its (n_query, n_instances) samples tensor of
+  /// indexed views into @p reference. @p sampleSize is a maximum — see
+  /// detail::draw_indices for the length contract. @p gen advances past the
+  /// draw's reserved seed block, so consecutive calls draw fresh samples.
+  template <typename T, typename FilterF, typename DistF>
+  SubsampleHandle<PointCloud<T>> sample_subsets(const PointCloud<T>& reference,
```
bwehlin: I'm not really understanding why we have this path. Can't we have one path that does the per-row subsampling and then reuse that as a fallback for small cases called directly as a loop from Python?

Anything large enough should not be using this since it results in very large weight matrices (unless I'm reading something wrong).

include/sbear/sampling/subsample_task.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+        const size_t nInstances = m_out.shape(1);
+        return parallel_walk_async(m_out,
+            [this, nInstances](const std::vector<size_t>& cellIdx) {
+          if (stop_requested())
+            return;
+
+          // cellIdx = (query, instance); the engine is seeded per cell from
+          // the reserved block, so draws are independent of thread count.
+          auto engine = m_seedBlock.sub_generator(cellIdx[0] * nInstances + cellIdx[1]);
```
bwehlin: Why are we pulling seed blocks manually here instead of letting parallel_walk handle it?

include/sbear/sampling/subsample_task.hpp
```diff
@@ -0,0 +1,112 @@
...
+          m_sampleSize(sampleSize), m_replace(replace), m_seedBlock(seedBlock)
+      { }
+
+    private:
+      tf::Future<void> run_async(Executor& exec) override
+      {
+        next_step(m_out.size(), "Drawing subsamples.", "subsample");
+
+        const size_t nInstances = m_out.shape(1);
+        return parallel_walk_async(m_out,
```
bwehlin: It's better to walk the input here. For the output we'd just take the cell index and tack on the extra dimension in the lambda body. This way, we can compute the filter and weights for each query point and then create all the subsamples for that query point.

include/sbear/sampling/subsample_task.hpp
```diff
@@ -0,0 +1,112 @@
...
+    /// subsample from its query's prepared weight row. Subsamples store only
+    /// the drawn indices, sharing the reference's data buffer.
+    template <typename ElemT>
+    class SubsampleTask : public StoppableTask<void>
+    {
+      using T = typename ElemT::value_type;
+      using SeedBlock = DefaultRandomGenerator::Block;
+
+    public:
+      SubsampleTask(ElemT source, Tensor<T> weights, Tensor<size_t> nEligible,
```
bwehlin: What is source? Why don't we keep query and reference as the names?

include/sbear/sampling/weighted_draw.hpp
```diff
@@ -0,0 +1,152 @@
...
+    T total = T(0);
+    for (size_t i = 0; i < weights.size(); ++i)
+    {
+      T w = weights[i];
+      if (w < T(0))
+        throw std::invalid_argument("sampling weights must be non-negative");
+      total += w;
+      cdf[i] = total;
+    }
+    if (!(total > T(0)))
```
bwehlin: Why not `if (total <= T(0))`?

include/sbear/sampling/weighted_draw.hpp
```diff
@@ -0,0 +1,152 @@
...
+  size_t prepare_weight_row(std::span<T> row, bool toCdf)
+  {
+    size_t nEligible = 0;
+    T total = T(0);
+    for (T& w : row)
+    {
+      // Reject negatives rather than counting them ineligible: an invalid
+      // row must not become a silent empty draw.
+      if (w < T(0))
+        throw std::invalid_argument("sampling weights must be non-negative");
```
bwehlin, line 66: Aren't we checking this already when preparing the CDF? If we are checking it here, could we just replace the CDF creation with `std::inclusive_scan`?

include/sbear/sampling/weighted_draw.hpp
```diff
@@ -0,0 +1,152 @@
...
+  template <typename T, typename EngineT>
+  Tensor<uint64_t> draw_without_replacement(std::span<const T> weights, size_t nDraws,
+                                            EngineT& engine)
+  {
+    std::vector<T> remaining(weights.begin(), weights.end());
+    std::vector<T> cdf;
+    Tensor<uint64_t> drawn({nDraws});
+    for (size_t drawIdx = 0; drawIdx < nDraws; ++drawIdx)
+    {
+      build_cdf(cdf, remaining);
```
bwehlin: I think this is correct, but maybe better to do weighted reservoir sampling instead? See https://utopia.duth.gr/~pefraimi/research/data/2007EncOfAlg.pdf

include/sbear/sampling/weighted_draw.hpp
```diff
@@ -0,0 +1,152 @@
...
+        : draw_without_replacement(row, std::min(sampleSize, nEligible), engine);
+  }
+
+  /// The contiguous row of one query point in a (n_query, n_reference)
+  /// weight matrix (prepare_weight_matrix normalizes contiguity once).
+  template <typename T>
+  std::span<T> weight_row(const Tensor<T>& weights, size_t queryIdx)
+  {
+    const size_t nReference = weights.shape(1);
+    return {weights.data() + queryIdx * nReference, nReference};
```
bwehlin: We are making assumptions on data layout here. Would it be better to use an indexing function on the tensor instead?

include/sbear/sampling/weighting.hpp
```diff
@@ -0,0 +1,197 @@
...
+  // way, code downstream of compute_weights only ever sees plain weights.
+  // ===========================================================================
+
+  /// Euclidean distance between query point @p queryIdx and reference point
+  /// @p refIdx. The samplers validate the shared dimension once up front, so
+  /// this hot-path operator stays branch-free.
+  template <typename T>
+  struct EuclideanDistance
+  {
+    T operator()(const PointCloud<T>& query, size_t queryIdx,
```
bwehlin: Rather than naming these query and reference, I would use x and y (or similar). This way, the function is more general in the sense that it is natural to apply in other cases where we might want to take distances between point clouds.

include/sbear/sampling/weighting.hpp
```diff
@@ -0,0 +1,197 @@
...
+  /// Weight 1 when the filter value lies in [low, high], 0 outside. With the
+  /// distance filter that is a disk (low = 0), an annulus, or the whole cloud
+  /// (the default, high = +inf). Member names match the Python spec.
+  template <typename T>
+  struct Uniform
+  {
+    T low = T(0);
+    T high = std::numeric_limits<T>::infinity();
+
+    T operator()(T v) const { return (v >= low && v <= high) ? T(1) : T(0); }
```
bwehlin: Let's put `[[nodiscard]]` on these operators. In most cases, they can also be made `noexcept`.

include/sbear/sampling/weighting.hpp
```diff
@@ -0,0 +1,197 @@
...
+  };
+
+  /// Unnormalized Gaussian exp(-((v - mean) / sigma)^2 / 2) of the filter
+  /// value, offering the log-space channel: a raw exp would underflow to an
+  /// artificial all-zero row once every reference point sits ~38 sigma
+  /// (float64; ~14 float32) from the mean, while the max-shifted exp samples
+  /// this exact Gaussian at any distance.
```
bwehlin: I don't understand this comment. To me it looks like the `struct` is just computing a Gaussian the normal "math" way?

include/sbear/sampling/weighting.hpp
```diff
@@ -0,0 +1,197 @@
...
+      if (!(maxLog > -std::numeric_limits<T>::infinity()))
+        maxLog = T(0);  // all -inf (empty region): avoid (-inf) - (-inf) = NaN
+      for (T& e : row)
+        e = std::exp(e - maxLog);
+    }
+
+    /// Max-shift every row of a freshly computed (hence contiguous)
+    /// log-weight matrix into plain weights, in parallel.
+    template <typename T>
+    void exponentiate_log_rows(Tensor<T>& weights, Executor& exec)
```
bwehlin: We should rework this to have the walk at a different level per earlier comment.

include/sbear/distance_matrix.hpp
```diff
@@ -24,6 +26,16 @@ namespace sb
   /// implicit zeros on the diagonal.
   /// For i != j, element (i, j) maps to storage index
   /// max(i,j)*(max(i,j)-1)/2 + min(i,j).
+  ///
+  /// A matrix may also be an *indexed view*: it shares another matrix's buffer
+  /// and selects a subset of points through an attached index set, so size() and
+  /// operator()(i, j) report the principal submatrix over those indices. This lets
+  /// a tensor of subsampled distance matrices store one shared source plus small
+  /// index arrays instead of re-storing every sub-matrix. Access is transparent,
+  /// so Ripser (which uses only size()/operator()) needs no special case. An
```
bwehlin: Ripser is not a "first-class citizen" in this context, so it shouldn't be part of the design contract.

include/sbear/distance_matrix.hpp
```diff
@@ -71,21 +83,59 @@ namespace sb
...
+    [[nodiscard]] const Tensor<uint64_t>& indices() const { return m_indices; }
+
+    /// Return an independent value. An owning matrix is deep-copied; an indexed
+    /// view keeps sharing the (immutable) source buffer but copies its index
+    /// array, so the result never aliases another cell's indices. Tensor cells
+    /// route stores through store_copy, which prefers copy().
     [[nodiscard]] DistanceMatrix copy() const
     {
+      if (is_indexed())
+        return DistanceMatrix(*this, m_indices.copy());
```
bwehlin: Let's make this function take a `bool keepSource = true` that does what is happening now, and when `keepSource == false`, we also copy the source buffer (so that we can get a completely clean slate on request). This should carry over to the Python interface and be documented.

include/sbear/distance_matrix.hpp
```diff
@@ -71,21 +83,59 @@ namespace sb
...
+    /// Number of points: selected indices when indexed, otherwise the full size.
+    [[nodiscard]] size_t size() const { return is_indexed() ? m_indices.shape(0) : m_size; }
+
+    /// Number of points in the (possibly shared) source buffer — equals size()
+    /// unless indexed. This is the matrix data() spans, used when serializing
+    /// the shared source of indexed views.
+    [[nodiscard]] size_t source_size() const { return m_size; }
+
+    /// Number of stored compressed entries (owning matrices only).
     [[nodiscard]] size_t storage_count() const { return storage_size(m_size); }
```
bwehlin: Arguably, this should return the size of the index array for the indexed case

include/sbear/distance_matrix.hpp
```diff
@@ -71,21 +83,59 @@ namespace sb
...
 
     [[nodiscard]] EntryProxy operator()(size_t i, size_t j)
     {
+      if (is_indexed())
+        throw std::logic_error("cannot write to an indexed distance-matrix view");
```
bwehlin: Let's do COW (copy-on-write) here instead, materializing and setting *this to the materialized version and then carrying on with the write. The indexing is a storage optimization, not behavior.

include/sbear/distance_matrix.hpp
```diff
@@ -146,7 +226,8 @@ namespace sb
     }
 
     std::shared_ptr<T[]> m_data;
-    size_t m_size;
+    size_t m_size;                 // size of the (shared) source buffer
+    Tensor<uint64_t> m_indices;    // rank-1 when an indexed view, empty otherwise
```
bwehlin: Possibly better class layout by going large to small (due to alignment): `m_indices`, `m_data`, `m_size`

include/sbear/point_cloud.hpp
```diff
@@ -6,11 +6,113 @@
...
+  ///
+  /// A PointCloud either owns its coordinates (like any Tensor<T>) or is an
+  /// indexed view: it shares another cloud's coordinate buffer and selects rows
+  /// through an attached index set. Access via n_points()/dim()/operator()(i, j) is
+  /// transparent to which mode it is in, so consumers (e.g. Ripser) need no
+  /// special case. This lets a tensor of subsamples store one shared source plus
+  /// small index arrays instead of re-storing every (possibly high-dimensional)
+  /// point.
+  template <ArithmeticType T>
+  class PointCloud : public Tensor<T>
```
bwehlin: This requires `Tensor<T>` to have a `virtual` destructor declared (`virtual ~Tensor() = default`) to avoid UB on deletion. Although, I'm not sure inheritance is the right hammer here since we are probably not intending to have runtime polymorphism between tensors and point clouds.

Let's rework before full review.

Let's make an interface `Indexable` in Python and have `Tensor` and `PointCloud` both inherit this (to enable X[1:4,0:1], etc...)

include/sbear/point_cloud.hpp
```diff
@@ -6,11 +6,113 @@
...
+  /**
+   * Cast a tensor of point clouds (Tensor<PointCloud<U>>) to a different precision
+   * (Tensor<PointCloud<T>>), converting each point cloud's coordinates.
+   */
+  template <typename T, typename U>
+  requires std::is_constructible_v<T, U>
+  [[nodiscard]] Tensor<PointCloud<T>> pcloud_cast(const Tensor<PointCloud<U>>& src)
+  {
+    Tensor<PointCloud<T>> result(src.shape());
+    walk(src, [&](const std::vector<size_t>& idx) {
```
bwehlin, line 188: Are we sure walk works with indexed tensors? If not, we should make it work.

We also need tests for this.

include/sbear/tensor.tpp
```diff
@@ -1704,6 +1693,27 @@ namespace sb
   bool allclose(const MatT& a, const MatT& b,
                 typename MatT::value_type atol, typename MatT::value_type rtol)
   {
+    // An indexed view's raw buffer is the full shared source, not its
+    // principal submatrix, so views compare elementwise (like operator==).
+    if constexpr (requires { a.is_indexed(); })
```
bwehlin: Higher-level functionality should not need to know about the indexing. Also we need tests for allclose for indexed cases.

src/python/sampling/py_subsample.cpp
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+  private:
+    // -------------------------------------------------------------------------
+    // Descriptor classes, nested under the precision namespace
+    // (Subsample32.Gaussian etc.). The descriptors are the core functors
+    // themselves, constructed member-by-member; parameter validation lives in
+    // the Python spec classes.
+    // -------------------------------------------------------------------------
+    static void register_descriptors(PyClass& cls)
```
bwehlin: "Descriptors" is a bit vague. Better to define two functions, register_filters, register_distributions

src/python/sampling/py_subsample.cpp
```diff
@@ -0,0 +1,194 @@
...
+    // Entry points: named static functions taking the functor variants. Each
+    // visitor holds the non-variant arguments; std::visit instantiates its
+    // call operator once per functor combination.
+    // -------------------------------------------------------------------------
+
+    struct SampleSubsetsCall
+    {
+      const TensorT& reference;
+      const TensorT& query;
+      size_t sampleSize;
```
bwehlin: Let's default initialize sampleSize, nInstances, replace and gen (nullptr).

src/python/sampling/py_subsample.cpp
```diff
@@ -0,0 +1,194 @@
...
+      py::tuple operator()(const DistF& distribution) const
+      {
+        // Same GIL story as SampleSubsetsCall above.
+        auto handle = [&] {
+          py::gil_scoped_release release;
+          return sb::sampling::sample_subsets_distmat(
+              source, query, distribution, sampleSize, nInstances, replace,
+              resolve_generator(gen), sb::default_executor());
+        }();
+        return to_tuple(std::move(handle));
```
bwehlin: It's a bit hard for me to understand what is going on here and whether the lifetime of `this` will outlive the lambda. Why can't we just spawn a `StoppableTask` instead of having an intermediate class that has a lambda and data?

src/python/sampling/py_subsample.cpp
```diff
@@ -0,0 +1,194 @@
...
+    static Gen& resolve_generator(Gen* gen)
+    {
+      return gen ? *gen : sb::default_generator();
+    }
+
+    /// Unpack a SubsampleHandle into the (task, samples) pair returned to Python.
+    template <typename ElemT>
+    static py::tuple to_tuple(sb::sampling::SubsampleHandle<ElemT> handle)
+    {
+      return py::make_tuple(std::move(handle.task), std::move(handle.samples));
```
bwehlin: Instead of doing this, we could, for example, declare a dummy tensor in the Python code and pass it to the subsampling for use as output. However, the pattern of having to return something other than void is common enough that we maybe should take a broader look on the architecture level to see if the flow can be improved. We do have some output code that works with outputs via tuples, so we could do the same thing, but we probably should avoid having these sorts of helper functions that are one-liners with a very generic name while they do something quite specific.

src/python/py_tensor.cpp
```diff
@@ -116,6 +117,22 @@ namespace
...
   }
 
+  // The element type of a PointCloud tensor. Registered as a subclass of the
+  // scalar Tensor<T> so it can be returned to Python; an element may be an indexed
+  // view sharing a source cloud (materialized lazily — see stablebear.tensor.PointCloud).
+  template <typename T>
+  void register_point_cloud_element(py::module_& m, const std::string& suffix)
+  {
+    using PC = sb::PointCloud<T>;
+    py::class_<PC, sb::Tensor<T>>(m, ("PointCloud" + suffix + "_inner").c_str())
```
bwehlin: Why _inner?

src/python/py_tensor.hpp
```diff
@@ -15,11 +15,25 @@
...
+#include <type_traits>
 
 namespace sb_py
 {
   void register_tensor_bindings(pybind11::module_& m);
 
+  // A point cloud element is any type derived from Tensor<scalar> (i.e. PointCloud<T>),
+  // which may be an indexed view and so must be materialized before crossing to Python.
+  template <typename T, typename = void>
+  struct is_point_cloud : std::false_type {};
```
bwehlin: These belong in the C++ layer rather than the Python layer.

stablebear/persistence/homology.py
```diff
@@ -112,6 +113,10 @@ def compute_persistent_homology(
         return out
 
     # --- Point cloud input path ---
+    if isinstance(X, PointCloud):
+        # A single (possibly indexed) cloud — materialize to a coordinate tensor.
+        X = X.materialize()
```
bwehlin: We shouldn't have to materialize these as long as `shape` and `operator()` do the right things for indexed point clouds.

stablebear/sampling/distributions.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class Uniform:
+    r"""Uniform weight over a band of filter values :math:`[\text{low}, \text{high}]`,
+
+    .. math::
+        D(v) = \begin{cases} 1 & \text{if } \text{low} \le v \le \text{high} \\
+                             0 & \text{otherwise.} \end{cases}
+
+    Applied to the Euclidean distance, this samples uniformly from a region
```
bwehlin, line 20: Maybe we can put this under an example section instead of the main copy?

stablebear/sampling/distributions.py
```diff
@@ -0,0 +1,102 @@
...
+    ----------
+    low : float, optional, keyword-only
+        Lower band edge :math:`\text{low} \ge 0`, by default 0.0.
+    high : float, optional, keyword-only
+        Upper band edge. If ``None`` (the default) it is :math:`+\infty`, so every
+        point at distance :math:`\ge` ``low`` is included. Must be strictly greater
+        than ``low``.
+    """
+
+    def __init__(self, *, low=0.0, high=None):
```
bwehlin: Let's do non-keyword-only. It's clear what is meant by Uniform(0, 1) etc.

stablebear/sampling/distributions.py
```diff
@@ -0,0 +1,102 @@
...
+        low = float(low)
+        if not (low >= 0):
+            raise ValueError("low must be non-negative.")
+        high = float("inf") if high is None else float(high)
+        if not (high > low):
+            raise ValueError("high must be strictly greater than low.")
+        self.low = low
+        self.high = high
+
+    def _descriptor(self, backend):
```
bwehlin: Maybe `_native` instead?

stablebear/sampling/distributions.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+
+class Gaussian:
+    r"""Unnormalized Gaussian of the filter value,
+
+    .. math::
+        D(v) = \exp\!\left(-\tfrac{1}{2}\left(\frac{v - \mu}{\sigma}\right)^2\right).
+
+    Applied to the Euclidean distance, this concentrates sampling probability on
```
bwehlin, line 55: Again, examples.

stablebear/sampling/distributions.py
```diff
@@ -0,0 +1,102 @@
...
+    Parameters
+    ----------
+    mean : float, optional, keyword-only
+        Center :math:`\mu`, by default 0.0. Must be finite.
+    sigma : float, optional, keyword-only
+        Standard deviation :math:`\sigma`, by default 1.0. Must be positive
+        and finite.
+    """
+
+    def __init__(self, *, mean=0.0, sigma=1.0):
```
bwehlin: I think we can do non-keyword here as well.

stablebear/sampling/subsample.py
```diff
@@ -0,0 +1,294 @@
...
+
+# The only accepted distributions. Each owns a fully fused C++ fast path
+# (distances + weighting + draw) for both point-cloud and distance-matrix
+# references; arbitrary callables are intentionally not supported.
+_BUILTIN_DISTRIBUTIONS = (Gaussian, Uniform)
+
+
+def _backend(dtype):
+    """The precision-specific C++ subsampling backend for a FloatTensor dtype."""
+    return cpp_samp.Subsample32 if dtype == float32 else cpp_samp.Subsample64
```
bwehlin: We have some established patterns for grabbing backends based on type (including checking for invalid types). We should use those here as well.

stablebear/sampling/subsample.py
```diff
@@ -0,0 +1,294 @@
...
+    raise ValueError(
+        f"reference is a square ({n_points}, {n_points}) array, which is "
+        "ambiguous: it could be a point cloud or a precomputed distance "
+        "matrix. Wrap it in stablebear.DistanceMatrix.from_dense(...) to "
+        "subsample by the stored distances, or in stablebear.FloatTensor(...) "
+        "to subsample its rows as points."
+    )
+
+
+def _query_is_indices(query):
```
bwehlin, line 111: I'm not sure why we have to pull out indices etc in the Python layer? Can we rework to have C++ work with the appropriate tensor types directly?

stablebear/base_tensor.py
```diff
@@ -260,6 +260,83 @@ def _to_py_tensor(self, data):
         return IntPcfTensor(data)
 
 
+class PointCloud:
```
bwehlin: I'm not sure base_tensor is the right place for this

test/python/sampling/test_subsample.py
```diff
@@ -0,0 +1,559 @@
...
+def test_reproducible_with_seed(float_kind):
+    np_float, _ = float_kind
+    R = np.random.default_rng(2).standard_normal((40, 2)).astype(np_float)
+    X = np.random.default_rng(3).standard_normal((3, 2)).astype(np_float)
+
+    a = subsample_relative(R, X, sample_size=10, n_instances=5, generator=sb.random.Generator(7))
+    b = subsample_relative(R, X, sample_size=10, n_instances=5, generator=sb.random.Generator(7))
+    c = subsample_relative(R, X, sample_size=10, n_instances=5, generator=sb.random.Generator(8))
+
+    for i in range(3):
```
bwehlin: Don't we have equality checks on point cloud tensors already?

test/python/sampling/test_subsample.py
```diff
@@ -0,0 +1,559 @@
...
+        el[0, 0] = 1.0
+
+    # A writable copy: mutating it leaves the source reference cloud untouched.
+    m = el.materialize()
+    m[0, 0] = 123.0
+    assert float(m[0, 0]) == 123.0
+    assert not np.any(R == 123.0)
+
+
+def test_owning_pointcloud_is_mutable():
```
bwehlin, line 431: We probably need to add a flag to point cloud tensors whether we have ever extracted an indexed view from it. If that's the case, before we mutate, we should make a copy of the source so that the indexed views do not change.

include/sbear/sampling/subsample_task.hpp
FilipNorth: The walk is over query points, but the reproducibility contract is per output cell: each query fans out into n_instances draws, and each draw needs its own engine so the result is independent of how the walk is scheduled across threads. The parallel_walk(gen) overload seeds one engine per walked cell — i.e. per query — which would make instance i's draw depend on the draws that ran before it on the same worker. So the block is reserved up front, sized to out.size() rather than the walked queries (in the synchronous prologue, which is also what advances the caller's generator so consecutive calls draw fresh samples), and sub-seeded by flat output index q * n_instances + i. There's now a comment in run_async spelling this out.

include/sbear/sampling/weighted_draw.hpp
FilipNorth: total <= T(0) lets a NaN total through (NaN comparisons are all false), and a single NaN weight makes the whole sum NaN — which would then feed straight into uniform_real_distribution. Writing it as !(total > T(0)) makes NaN fail validation too. It's also gated on nEligible > 0 so an all-zero row still counts as a valid empty region rather than an error

include/sbear/sampling/weighted_draw.hpp
FilipNorth: After the rework this is all one pass: prepare_weight_row does the negative/NaN validation, counts the eligible entries, and (with replacement) builds the prefix sums in place in a single loop, the duplicate check is gone. It can't be a plain std::inclusive_scan because the same pass has to compute nEligible and reject negatives, and without replacement the row must stay raw weights for the E-S draw, so the CDF conversion is conditional.

include/sbear/sampling/weighting.hpp
FilipNorth: the struct itself is just the textbook Gaussian. What the comment was (badly) trying to say is why log_weight exists alongside operator(): the sampler evaluates a query's row in log-space with a max-shift, because the raw exp underflows to an artificially all-zero row once every reference point sits ~38σ from the mean (float64; ~14σ in float32), silently turning a valid region into an empty draw. The comment is rewritten to say exactly that.

include/sbear/sampling/weighted_draw.hpp
FilipNorth: Draws without replacement now use Efraimidis–Spirakis weighted reservoir sampling

include/sbear/sampling/weighted_draw.hpp
FilipNorth: The full weight matrix (whose row-major layout was being assumed) no longer exists; drawing operates on a per-query row std::span prepared in place, so there is no cross-row layout assumption left.

src/python/sampling/py_subsample.cpp
FilipNorth: the C++ entry points now return a self-contained SubsampleHandle{task, samples}, and the binding unpacks it with a small documented to_tuple into the (task, samples) tuple, the tuple-output pattern we already use elsewhere, and the generically-named helper is gone. Agreed that the task-output flow deserves a broader architecture look, but I'd keep that out of this PR's scope.

src/python/py_tensor.hpp
FilipNorth: The is_point_cloud trait moved to the C++ layer (point_cloud.hpp:144-151); the binding layer uses sb::is_point_cloud_v.

test/python/sampling/test_subsample_distmat.py
```diff
@@ -0,0 +1,297 @@
...
+                             generator=sb.random.Generator(4))
+    pos = subsample_relative(dm, np.array([24, 0]), sample_size=5, n_instances=3,
+                             generator=sb.random.Generator(4))
+    for i in range(2):
+        for j in range(3):
+            assert np.array_equal(np.asarray(neg[i, j].indices),
+                                  np.asarray(pos[i, j].indices))
+
+
+def test_distmat_gaussian_far_shell_samples_nearest_not_empty():
```
MattiasBjerklov, line 100: I think that this is fine, but perhaps worth checking. The test has a (very low) probability of failing if another seed is chosen.

PR #217, opened 2026-07-16
Homological kernel

FilipNorth:
# Add the 0th homological kernel

## Feedback addressed

For now I think we should postpone adding BOOST for the MST. The code surruonding it is more complicated than the algorithm itself right now.

- **Simplified Python API** Removed the options for built in transformations on a single pointcloud. Now users must supply 2 pointclouds or 2 distmats. Added homology dimension option but limit to 0 for now. 
- **MST extracted to a shared algorithms header** (`sb::mst_merge_order`,
  `sb::MergeEdge`, `sb::SquaredEuclideanDistance` in
  `algorithms/minimum_spanning_tree.hpp`); the kernel header keeps only the
  kernel-specific cross-filtration.
- **Boost.Graph was evaluated and declined for now.** A PoC (implicit
  complete-graph adaptor + `prim_minimum_spanning_tree`, Boost 1.88) produced
  identical results and ran ~10% faster — but the adaptor is ~70 lines of
  hand-written BGL protocol plumbing replacing ~45 lines of textbook Prim, and
  wheel builds would need a ~656 MB Boost FetchContent in every CI job (system
  Boost is not viable under cibuildwheel).

## Goal

Given two metrics on the same point set with `d' ≤ d` pointwise (e.g. a point
cloud and its projection), compute the persistence barcode of the kernel of the
induced map `μ: H₀(VR(X, d)) → H₀(VR(X, d'))`. The kernel barcode measures how
much 0-dimensional structure the smaller metric collapses.

The reference implementation (paper/thesis) was Python + numba, hard-coded to
2D clouds with the diagonal projection. This PR generalizes it to:

- any dimension and any metric — point clouds (Euclidean, no distance matrix
  ever materialized) and precomputed distance matrices,
- batched tensors with instance-level parallelism (the real workload:
  hundreds of thousands of small instances, n ≈ 10–50),
- both float32 and float64, with the usual tensor-in/tensor-out API.

## What this PR does

**Algorithm.** Two array-Prim MST passes (O(n²) distance evaluations, O(n)
memory) produce the d'-merge order (the births) and the d-MST (the death
timeline). Deaths are answered by a per-merge Kruskal sweep: seed a union-find
with the current d'-components, add d-MST edges ascending, stop when the
merging pair connects. All sweeps together cost exactly n(n−1)/2 bounded
union-find operations — O(n²·α(n)) time, O(n) memory; the O(n²) MSTs dominate
end-to-end.

The MST and the on-demand squared-Euclidean distance oracle live in
`include/sbear/algorithms/minimum_spanning_tree.hpp` as general primitives
(namespace `sb`), not in the kernel — they are the exact machinery a future
single-linkage H₀ path needs to bypass Ripser for `dim=0` persistence.

**API.**

```python
def compute_homological_kernel(
    X: PointCloudTensor
    | DistanceMatrix
    | DistanceMatrixTensor
    | FloatTensor
    | np.ndarray,
    Y: ...,                 # same kind and shape as X
    *,
    dim: int = 0,           # reserved for higher degrees; only 0 supported
    verbose: bool = False,
) -> BarcodeTensor:

bcs = pers.compute_homological_kernel(X, Y)   # X carries d, Y carries d'
```

`X` carries the larger metric `d`, `Y` the dominated `d'`. Both must be the
same kind and shape; element `i` of `X` pairs with element `i` of `Y`. Point
clouds always use the Euclidean metric and stay in ambient dimension (a
projection is expressed in the original coordinates); for any other metric,
pass precomputed distance matrices.

Earlier revisions offered named `transform=` presets ("diagonal",
"coordinate") and a `metric=` parameter. Both were removed: there is now
exactly one way to call the function, the `d'` source is always visible at the
call site, and no preset-certification machinery needs to exist. Domination is
enforced at runtime for every input instead of being proven per-preset.

Naming convention across the boundary: Python speaks in data (`X`, `Y` —
arrays, clouds), C++ speaks in math (`d`, `d'` — the metrics they induce); the
mapping is documented once at the pybind seam.

**Robustness.** Domination is enforced (death below birth ⇒ throw) with a
`sqrt(eps)`-relative tolerance: pure roundoff clamps to an empty bar, genuine
violations throw. Every instance with n points yields exactly n−1 finite bars;
malformed elements throw with the failing index in the message.

**Performance.** The point-cloud oracle computes squared distances (MST
construction only compares, and x → x² is monotone), moving the sqrt from the
O(n²) inner loop to the 2(n−1) stored merge distances. Batch throughput:
~3–52 µs per instance at n = 10–50 single-threaded (~17k instances/s/core at
n = 50), parallelized across instances. A single n = 20k instance runs in
~5–7 s, dominated by the two MST passes.

## What worked well

- **The code follows the Ripser pipeline in its structure** — task class,
  type dispatch in the task layer, per-instance wrappers, `parallel_walk`
  batching — deviating only in how the barcodes themselves are computed. New
  code slots into the pre-aligned structure instead of inventing its own.
- **The distance-oracle pattern.** The core is templated on one `DistT` with
  `operator()(i, j)`; distance matrices already are oracles, point clouds get
  an on-demand (squared-)Euclidean one. No materialized matrices, and every
  stage is unit-testable with 4-point fixtures.
- **The MST-multiset invariant** (births = d'-MST weights, deaths a
  permutation of d-MST weights) turned a deep correctness property into a
  cheap, exact test that needs no reference implementation.

## Verification

The test suite is deliberately non-self-referential:

- hand-computed fixtures for each phenomenon (merge-order inversion, tie
  invariance, coincident projections, deaths carried by non-merged pairs,
  contraction-shortcut and staircase cases),
- the exact MST-multiset invariant on random inputs, compared bitwise,
- randomized cross-checks of both input routes against an independent O(n³)
  dense reference implementation (explicit quotient-ultrametric maintenance),
- route/dtype consistency (pcloud vs distmat, float32), batch semantics, and
  all error paths.

## What could be improved

- **Large single instances run on one core.** Only the batch dimension is
  parallelized. The two MST passes are independent and each Prim iteration is
  data-parallel.
- **O(n²) is not optimal for Euclidean point clouds.** An exact Euclidean MST
  via dual-tree Borůvka (KD-tree) is ~O(n log n) and would cut large-cloud
  times by orders of magnitude; the distmat path is already optimal (any exact
  MST must read Ω(n²) entries).
- **Approximate MSTs** could be a future option for much faster construction
  on large clouds.

## Future goals

- **Single-linkage H₀ without Ripser.** The extracted MST is the direct route
  to `dim=0` persistence for point clouds (deaths = MST edge weights, no
  distance matrix materialized) — the follow-up that motivated the extraction.
- **Higher degrees (p ≥ 1).** This is the degree-0 case of kernel persistence
  (Cohen-Steiner–Edelsbrunner–Harer–Morozov, SODA 2009). Rank–nullity gives
  the kernel Betti curves in every degree as (ordinary persistence of d) −
  (image persistence of the inclusion), and image persistence for exactly this
  setting is solved (Bauer–Schmahl, SoCG 2022 / ripser-image). Full
  higher-degree pairing is a research problem (no union-find analogue above
  H₀).
- **GPU batch path.** The dense small-n workload is warp-per-instance shaped;
  fits `src/cuda/` if the subsampling volumes demand it.
- **Projection helpers.** Small utilities that build `Y` from `X` (diagonal,
  coordinate, PCA projections) — as plain functions producing point clouds,
  not API modes; integration with the subsampling branch.
- **External ground truth**: a Dionysus (or ripser-image) cross-validation
  harness would validate against code sharing no ancestry with this
  implementation.


docs/homological-kernel.rst
```diff
@@ -0,0 +1,231 @@
...
+  under the coarse distance :math:`d` -- the moment the discrepancy is resolved.
+
+Since :math:`d' \le d`, every death is at least its birth, so all bars are
+well-formed. A **zero-length bar** :math:`[w_i, w_i)` means the two distances
+merge those components at the very same scale: no disagreement there. Long bars
+mark where the two structures genuinely diverge. A set of :math:`n` points always
+produces :math:`n - 1` bars, one per merge in the hierarchy.
+
+
+How the kernel is computed
```
bwehlin: Belongs in the internals docs

docs/homological-kernel.rst
```diff
@@ -0,0 +1,231 @@
...
+correlation case. The simplest way to read the output is as 0-th homology
+mimicking 1-st homology: it exposes the difference in topology between two
+distance spaces built on the same points.
+
+The generalization, and the proof of correctness for the algorithm below, come
+from forthcoming work by the authors -- an approved master's thesis that is not
+yet published.
+
+
+What the kernel measures
```
bwehlin: Too complicated for general users that may not be mathematicians. Can we add some pictures etc?

Add as a section on "Mathematical background" instead.

docs/homological-kernel.rst
```diff
@@ -0,0 +1,231 @@
+===========================
+The 0-th Homological Kernel
+===========================
+
+This guide covers how to use the 0-th homological kernel in stablebear. The
+method takes in point clouds or distance matrices and returns persistence
+barcodes, which can then be used in the normal persistent homology pipeline
+(see :doc:`persistence`).
+
+Background
```
bwehlin, line 16: Instead of the historical background, let's replace this section with what the thing actually does for end users, and add a citation to the master's thesis (even if it's not on DiVA yet)

docs/homological-kernel.rst
```diff
@@ -0,0 +1,231 @@
+===========================
+The 0-th Homological Kernel
```
bwehlin: Change to "Homological Kernels" and mention that only the H0 one is computed at the moment in the text.

include/sbear/algorithms/minimum_spanning_tree.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+  template <typename T>
+  struct MergeEdge
+  {
+    size_t a; // endpoints of the MST edge = representatives of the two merging components
+    size_t b;
+    T mergeDist;
+  };
+
+  // Distance type is anything answering d(i, j) on demand: in practice a
```
bwehlin: This can be expressed using concepts instead directly in the code.

include/sbear/algorithms/minimum_spanning_tree.hpp
```diff
@@ -0,0 +1,130 @@
...
+    merges.clear();
+    if (n <= 1)
+    {
+      return;
+    }
+    merges.reserve(n - 1);
+
+    std::vector<T> minDist(n, std::numeric_limits<T>::infinity());
+    std::vector<size_t> parent(n, 0);
+    std::vector<char> visited(n, 0);
```
bwehlin: Why not `vector<bool>`?

include/sbear/algorithms/minimum_spanning_tree.hpp
```diff
@@ -0,0 +1,130 @@
...
+      for (size_t j = 0; j < n; ++j)
+      {
+        if (!visited[j] && (!found || minDist[j] < minDist[nearest]))
+        {
+          nearest = j;
+          found = true;
+        }
+      }
+
+      visited[nearest] = 1;
```
bwehlin: use `bool` for true/false values

include/sbear/algorithms/minimum_spanning_tree.hpp
```diff
@@ -0,0 +1,130 @@
...
+  };
+
+  // Distance type is anything answering d(i, j) on demand: in practice a
+  // distance matrix, or an oracle such as SquaredEuclideanDistance that
+  // computes distances from a point cloud without materializing a matrix.
+  //
+  // Fills merges with the n-1 single-linkage merges of the metric, sorted ascending
+  // by merge distance. Array-based Prim: O(n^2) distance queries, O(n) memory.
+  template <typename DistT, typename T>
+  void mst_merge_order(const DistT &dist, size_t n, std::vector<MergeEdge<T>> &merges)
```
bwehlin: Why do we need `n` here? Can it be inferred from `dist`? Generally, we try to avoid having C style interfaces with `data` + `length`

include/sbear/persistence/compute_homological_kernel.hpp
```diff
@@ -0,0 +1,276 @@
...
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace sb::ph
+{
+  namespace detail
+  {
+    // Union-find root lookup with path halving.
+    inline size_t uf_find_root(std::vector<size_t> &ufParent, size_t x)
```
bwehlin: Why don't we create a class for union find instead? It looks like we're implementing union-find inline into another algorithm later. Let's have a reusable component instead -- this also makes the readability better.

include/sbear/persistence/compute_homological_kernel.hpp
```diff
@@ -0,0 +1,276 @@
...
+    // Full kernel computation for one instance: d' merge order (the births),
+    // d-MST (the death timeline), cross filtration. Requires d' <= d pointwise;
+    // cross_filtration throws otherwise. For n <= 1 the kernel barcode is empty.
+    // mergeDistTransform maps each stored merge distance from the scale the
+    // DistT functor computes in to the scale of the output bars (sqrt for the
+    // squared-Euclidean oracle, identity for oracles already in bar scale); it
+    // must be monotone so the merge order is unchanged.
+    template <typename DistT, typename T, typename PostF = std::identity>
+    void homological_kernel_single_impl(
+        const DistT &dDist, const DistT &dPrimeDist, size_t n, Barcode<T> &ret, PostF mergeDistTransform = {})
```
bwehlin: Same note about C style interfaces as earlier

include/sbear/persistence/compute_homological_kernel.hpp
```diff
@@ -0,0 +1,276 @@
...
+    // the current d'-components, add d-MST edges in ascending order, and stop
+    // as soon as a_i ~ b_i; the weight of the connecting edge is the death.
+    //
+    // The sweep for merge i stops after rank(v_i) edges, and the deaths are a
+    // permutation of the d-MST edge ranks, so all sweeps together perform
+    // exactly n(n-1)/2 bounded union-find operations plus an O(n) seed copy per
+    // merge: O(n^2 alpha(n)) time, O(n) memory. Throws if some death lands
+    // below its birth (beyond roundoff), which means d does not dominate d'.
+    template <typename T>
+    void cross_filtration(
```
bwehlin: `compute_cross_filtration`?

include/sbear/persistence/compute_homological_kernel.hpp
```diff
@@ -0,0 +1,276 @@
...
+        throw std::runtime_error(
+            "homological kernel: point cloud at index " + index_to_string(index) + " has unexpected shape " +
+            shape_to_string(pc.shape()) + " (should be (m, n))");
+      }
+
+      // Every rank-2 cloud flows through: 0 or 1 points give an empty barcode
+      // via the n <= 1 early-outs, and a zero-dimensional cloud (n, 0) induces
+      // the all-zero metric, yielding n-1 zero-length bars exactly like the
+      // distance-matrix route does for the equivalent all-zero matrix.
+      SquaredEuclideanDistance<T> dDist(pc); // captures pointer+strides, computes d^2 on demand
```
bwehlin, line 210: I'm starting to think that point clouds should carry a distance so that we can unify things later. This way, we can create distance spaces directly by wrapping a point cloud in a distance.

stablebear/persistence/homological_kernel.py
```diff
@@ -0,0 +1,168 @@
...
+    Y: PointCloudTensor
+    | DistanceMatrix
+    | DistanceMatrixTensor
+    | FloatTensor
+    | np.ndarray,
+    *,
+    dim: int = 0,
+    verbose: bool = False,
+) -> BarcodeTensor:
+    r"""Compute the homological kernel between two distance structures.
```
bwehlin: Should cite thesis.

This description is a bit too technical for non-mathematicians, especially "distance structures".

stablebear/persistence/homological_kernel.py
```diff
@@ -0,0 +1,168 @@
...
+
+    if dim < 0:
+        raise ValueError(f"dim must be non-negative (got {dim})")
+    if dim > 0:
+        raise NotImplementedError(f"only dim=0 is currently supported (got dim={dim})")
+
+    X = _normalize_kernel_input(X, "X")
+    Y = _normalize_kernel_input(Y, "Y")
+
+    if isinstance(X, PointCloudTensor) != isinstance(Y, PointCloudTensor):
```
bwehlin: This type checking is a bit fragile. What if both `X` and `Y` are `int` for example? Also, we have some established patterns for checking types elsewhere.

test/python/persistence/test_homological_kernel.py
```diff
@@ -0,0 +1,358 @@
+"""Whole-pipeline tests for compute_homological_kernel through the public API.
```
bwehlin, line 1: I will check the tests after the docs have been updated.

stablebear/persistence/homological_kernel.py
FilipNorth: Changed to distance matrixes or pointclouds. Unsure if we need to change "Compute the homological kernel between" as well.

docs/homological-kernel.rst
```diff
@@ -0,0 +1,418 @@
...
+merge identically, i.e. the cloud already lies on the diagonal. Larger scores
+mark stronger deviation from it. Comparing, averaging, or classifying whole
+cohorts of homological kernels works exactly as for any other barcode
+tensor -- see :doc:`persistence`.
+
+
+A worked example: Anscombe's quartet
+====================================
+
+Anscombe's quartet is four small datasets that share, to two decimals, the
```
bwehlin: Cite Anscombe

docs/homological-kernel.rst
```diff
@@ -0,0 +1,418 @@
...
+same mean, the same variance, the same Pearson correlation (:math:`r` = +0.82)
+and the same regression line, yet look nothing alike. It is the standard
```
bwehlin: "nothing alike" -> "quite different". "the standard" -> "a standard"

docs/homological-kernel.rst
```diff
@@ -0,0 +1,418 @@
...
+demonstration that summary statistics miss what a plot shows immediately.
```
bwehlin: summary statistics *can* miss

docs/homological-kernel.rst
```diff
@@ -0,0 +1,418 @@
...
+
+The whole computation -- project onto the diagonal, compute the homological
+kernel, summarise as a stable rank, integrate to a score -- fits in a dozen
```
bwehlin: summarise -> summarize

docs/homological-kernel.rst
```diff
@@ -0,0 +1,418 @@
+===================
+Homological Kernels
+===================
+
+This guide covers the homological kernel in stablebear: what it computes, how
+to prepare its paired inputs, and how to read the result. The kernel method
```
bwehlin: "kernel method" has established meaning in different context

docs/homological-kernel.rst
```diff
@@ -0,0 +1,418 @@
...
+returns ordinary persistence barcodes, so everything downstream of the
```
bwehlin: downstream *from*?

docs/homological-kernel.rst
```diff
@@ -0,0 +1,418 @@
...
+exist in every homology degree, but at the moment only the 0-th one
+(``dim=0``) is computed.
+
+
+Background
+==========
+
+The **homological kernel** compares two distances on the same set of points.
+The input is a point set together with two distances :math:`d` and :math:`d'`
+with :math:`d' \le d` everywhere -- typically a point cloud with its Euclidean
```
bwehlin: typically -> for example,

Maybe include a figure showing how the distances shrink.

docs/homological-kernel.rst
```diff
@@ -0,0 +1,418 @@
...
+distance, and a projection of the same cloud, which can only move points
+closer together. As the scale increases, each distance merges the points into
```
bwehlin: Unclear what "the scale increases" means -- is this the filtration scale?

The merging stuff needs a figure to be clear.

docs/homological-kernel.rst
```diff
@@ -0,0 +1,418 @@
...
+larger and larger clusters, and since :math:`d'` is the smaller distance, it
+merges them earlier. The homological kernel records those disagreements.
+
+The output is a **persistence barcode** with one bar per merge: the bar is
+born at the scale where two clusters merge under :math:`d'` and dies at the
+scale where the same clusters merge under :math:`d`. Where the two distances
+agree, the bars have length zero -- the more they disagree, the longer the
+bars.
+
+The main application is using it as a **correlation method**: pair two
```
bwehlin: it?

docs/homological-kernel.rst
```diff
@@ -0,0 +1,418 @@
...
+     - 14.396
+
+The two homology columns are identical to every decimal, because a rotated
+cloud is the same metric space. The homological kernel score moves with the
+correlation, because the diagonal stays fixed while the data turns relative
+to it. That fixed second structure is exactly what a single barcode lacks,
+and supplying it is what the homological kernel is for.
+
+
+Mathematical background
```
bwehlin, line 420: I would refer to this section in the beginning saying something like "for the more mathematically inclined reader, we refer to..."

include/sbear/algorithms/union_find.hpp
```diff
@@ -0,0 +1,54 @@
...
+  class UnionFind
+  {
+  public:
+    /// n singleton components.
+    explicit UnionFind(size_t n) : m_parent(n)
+    {
+      std::iota(m_parent.begin(), m_parent.end(), size_t{0});
+    }
+
+    /// Root of x's component, halving the path along the way.
```
bwehlin: Maybe point out that this is "path compression" (standard)

include/sbear/algorithms/union_find.hpp
```diff
@@ -0,0 +1,54 @@
...
+  /// unite() is directed: the caller passes the two roots and picks which one
+  /// survives. This is load-bearing for callers that cache roots across
+  /// unions (e.g. the homological-kernel cross filtration): a cached root can
+  /// only stop being a root when a union writes over its slot, so knowing
+  /// which root was absorbed lets the caller patch its caches exactly.
+  ///
+  /// Copy assignment reuses the target's capacity (plain vector copy
+  /// assignment), so repeatedly re-seeding a scratch instance from a fixed
+  /// seed allocates nothing in the steady state.
+  class UnionFind
```
bwehlin, line 22: I would name this data structure `DisjointSets`

docs/homological-kernel.rst
```diff
@@ -0,0 +1,418 @@
...
+dataset's distant outlier rather than anything about the relationship between
+the two variables. :math:`H_1` is nearly silent -- only dataset I has any
+loops at all. The other three appear in the middle panel's legend but have no
+visible curve, because their :math:`H_1` barcodes are empty and their stable
+ranks are flat at zero. Eleven points rarely enclose a hole.
+
+There is a deeper reason a plain barcode cannot measure correlation:
+:math:`H_0` and :math:`H_1` depend only on the distances between points, so
+**rotating a cloud cannot change them at all** -- while rotating a cloud
+changes its correlation completely. Rotating dataset I about its centre::
```
bwehlin: centre -> center

include/sbear/persistence/compute_homological_kernel.hpp
```diff
@@ -0,0 +1,272 @@
...
+            "homological kernel: input tensors must have the same shape (got " + shape_to_string(m_input.shape()) +
+            " and " + shape_to_string(m_inputPrime.shape()) + ")");
+      }
+
+      // Validate before touching m_ret so a failed spawn leaves the caller's
+      // out tensor intact.
+      m_ret = Tensor<Barcode<T>>(m_input.shape());
+
+      next_step(
+          m_input.size(), "Computing 0th homological kernel",
```
bwehlin: "H0 kernel" better than "0th homological kernel"?

